PR #10, opened 2018-10-23
Fix error

mbogomazov:
Fix error "MDTextField has no attribute '_cursor_blink'"


HeaTTheatR:
It's not a mistake

mbogomazov:
No, It is. Android app crashes if code contains '_cursor_blink'(if 'cursor_blink' all works correctly') and in original KivyMD library this attribute has name 'cursor_blink'.

PR #65, opened 2019-05-21
Semi-automated way to update icons

payala:
I realized that the materialdesignicons were a bit outdated, so I created a semi-automated way to update them.

The way the update works is that you need to download the webfont .zip package from https://materialdesignicons.com/

From that zip, we take two files:

* materialdesignicons-webfont.ttf: this file is just placed in `kivymd/fonts/` dir, replacing the old one.
* preview.html: this file is placed in `kivymd/utils/res/` dir.

Next step is to run the script `kivymd/utils/parse_icons.py`, which will read `preview.html` and generate a new `kivymd/icondefinitions.py` file.

If theres something I'm overlooking or this doesn't make much sense, please let me know.


HeaTTheatR:
@payala And where is the script for downloading the html file?

HeaTTheatR:
@payala Or is it a one-time update?

payala:
That's done manually atm, that's why it's not fully automated. Downloading the zip file and placing the .html and .ttf from the zip file in its place is manual.
Generating the icondefinitons.py file is what is automated.

So yeah, it's more like a one time update thing.

It would not be very difficult to fully automate it. But I don't have much time, so I thought it's better to push this today, and we can refine it later.

payala:
I don't know how the icondefinitions.py was created in the first place. Maybe there is already a way to generate it and I'm missing it.

HeaTTheatR:
@payala Then just make a new pool with updated materialdesignicons-webfont.ttf and icondefinitions.py file.

payala:
Ok, but if I leave out the parse_icons.py file and someone wants to update the icons in the future, he would have to find a way to update the icondefinitions.py file.

HeaTTheatR:
@payala I can do it. But it should be as automated as possible and run as a single command.

payala:
Ok, I will push only those two files and make it a one-time only update.

payala:
I think it's done now.

payala:
ok, will create another PR.

payala:
Sorry, I didn't see that you merged it.

PR #66, opened 2019-06-03
MDAccordionListItem improvements.

shadr:
1. When you click on accordionlistitem with unfinished animation, it throw exception. Here is the fix.
2. Changed `BoxLayout` to `StackLayout` for more smooth animation.
But, unfortunately, the second commit causes a new bug, a solution for which I haven't found yet. The bug is that the title starts shaking when you click on the widget.

### DIFF
--- a/kivymd/accordionlistitem.py
+++ b/kivymd/accordionlistitem.py
@@ -210,6 +210,7 @@ class MDAccordionListItem(BoxLayout):
     content = ObjectProperty()
     icon = StringProperty()
     title = StringProperty()
+    canAnim = True
 
     def check_open_box(self, instance):
         press_current_item = False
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@Shadr I do not observe any errors.

shadr:
Oh, that's my mistake. An exception is thrown if you click on the MDAccordionListItem when another MDAccordionListItem hasn't finished the animation. I tested on an example from accordionlistitem.py .

HeaTTheatR:
@Shadr Show me how you can make it ...

shadr:
Maybe I wrong and exception is thrown because of something else. I think I can close PR and just create Issue because my PR now do almost nothing.
https://youtu.be/sbUrKYHNo-c

HeaTTheatR:
@Shadr What you are showing is not playing on my computer.

PR #72, opened 2019-06-19
Adding parameter to set non active button color

GoBig87:
When adding a swipe manager to a white background all the rectangles disappear because they are hard coded to white.  Adding a parameter to set the non active rectangles so they can be adjusted for background colors.  This allows users to set the color like this 

```
        Builder.load_string(activity)
        start_screen = MySwiperManager(util=self.util)
        self.swiper_manager = start_screen.ids.swiper_manager
        paginator = MDSwiperPagination()
        paginator.color_round_not_active = [0, 0, 0, .2] # <- Color can be set here now
        paginator.screens = self.swiper_manager.screen_names
        paginator.manager = self.swiper_manager
        self.swiper_manager.paginator = paginator
        start_screen.add_widget(paginator)

        return start_screen
```

### DIFF
--- a/kivymd/managerswiper.py
+++ b/kivymd/managerswiper.py
@@ -232,10 +232,12 @@ class MDSwiperPagination(ThemableBehavior, BoxLayout):
     screens = ListProperty()
     items_round_paginator = []
     manager = ObjectProperty()
+    color_round_not_active = ListProperty([1, 1, 1, 1])
 
     def on_screens(self, instance, screen_names):
         for i, screen_name in enumerate(screen_names):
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
```python
class ItemPagination(ThemableBehavior, Widget):
    current_index = NumericProperty(0)
    color_round_not_active = ListProperty([1, 1, 1, 1])
```

HeaTTheatR:
@GoBig87 The color of the circles is changeable.

HeaTTheatR:
@GoBig87 I meant it.

PR #76, opened 2019-07-20
Add markup support for buttons

svareille:
Add possibility to use text markup inside buttons.

### DIFF
--- a/kivymd/button.py
+++ b/kivymd/button.py
@@ -232,6 +232,7 @@ Builder.load_string(
     width: lbl_txt.texture_size[0] + root.increment_width
     padding: (dp(8), 0)
     theme_text_color: 'Primary'
+    markup: False
 
     MDLabel:
         id: lbl_txt
@@ -243,6 +244,7 @@ Builder.load_string(
         height: self.texture_size[1]
         theme_text_color: root.theme_text_color
         text_color: root.text_color
+        markup: root.markup
         disabled: root.disabled
         valign: 'middle'
         halign: 'center'
@@ -313,6 +315,7 @@ Builder.load_string(
 
     size_hint_x: None
     width: dp(150)
+    markup: False
 
     BoxLayout:
         spacing: dp(10)
@@ -333,11 +336,13 @@ Builder.load_string(
             shorten: True
             theme_text_color: 'Custom'
             text_color: root.theme_cls.primary_color
+            markup: root.markup
 
 
 <MDRoundFlatIconButton>
     size_hint_x: None
     width: dp(150)
+    markup: False
 
     BoxLayout:
         spacing: dp(10)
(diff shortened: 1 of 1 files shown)


HeaTTheatR (commented):
It seems to me that it does not work.

svareille:
The default behaviour of kivy's labels is to have text markup disabled. These buttons act the same way.
In fact you have to set "markup: True" inside your button definition to use markup: 
``` Python
MDFillRoundFlatButton:
    text: "1[sup]st[/sup] button"
    markup: True
    elevation_normal: 2
    opposite_colors: True
    size_hint: 0.8, 0.8
    pos_hint: {'center_x': 0.5, 'center_y': 0.5}
```

HeaTTheatR (commented):
You don't even have a markup parametr - https://github.com/HeaTTheatR/KivyMD/pull/76/commits/9d69f5dc5357f76d7f9be3e1bc55ffac3a2ad438#diff-388bedbbd6a85396a0abaa4e81d04c19R366
It just won't work

svareille:
`MDLabel` inherit from kivy's `Label` so `markup` is avaliable.
I tested it before submitting: at least `MDFillRoundFlatButton`, `MDRoundFlatIconButton` and `MDRectangleFlatIconButton` works with markup.

HeaTTheatR (commented):
I think it would not be bad to make these changes for all classes of buttons.

svareille:
I didn't inspect the full tree, but I think that every button inherit from at least one of these classes. I will do some tests.

HeaTTheatR:
@SimonVareille I just do not have time to check it all on my own. I am also working on a lot of changes and new classes.

svareille:
Ok, I will test it for every buttons ant let you know.

svareille:
It works with every buttons:
image

Just by adding `markup: True` and the tag we want (modification of the 'buttons' section in `screens.py` for the kitchen_sink example):
```Python
MDFlatButton:
    text: 'MD[u]Flat[/u]Button'
    pos_hint: {'center_x': .5}
    markup: True

 MDRaisedButton:
    text: "MDRaised[sub]Button[/sub]"
    elevation_normal: 2
    opposite_colors: True
    pos_hint: {'center_x': .5}
    markup: True

MDRectangleFlatButton:
    text: "MDRectangle[b]Flat[/b]Button"
    pos_hint: {'center_x': .5}
    markup: True

...
```
(full code [here](https://gist.github.com/SimonVareille/ef99db3364e06579ec84ab01f439dd57))

HeaTTheatR:
@SimonVareille You ran tests with the current pool?

svareille:
I ran tests with the version of `kivymd/button.py` I propose, witch is based on your current master branch.

PR #103, opened 2019-09-06
Allow override starting coordinates for MDDropdownMenu

ahackit:
### Description of Changes
* MDDropdown has hard-coded  Start Coordinates to center_x, center_y of caller. 
* Added OptionProperty to allow switch based on whether user wants to grow based on center of caller or bottom of caller.

### Screenshots
Add images to explain us your changes. Paste urls here.

image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ demos/kitchen_sink/bin
 # Editors
 .idea
 *.swp
+.vscode/settings.json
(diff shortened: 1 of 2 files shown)


HeaTTheatR:
This is already implemented in version 0.101.0

Снимок экрана 2019-09-06 в 20 44 54

ezgif-5-9006fe794cf3

PR #104, opened 2019-09-07
Load the spinner with MDSpinner().

ghost:
### Description of Changes
* Whenever you add the spinner as a class (for example, with ```self.add_widget(MDSpinner())```, it throws the error ```"IndexError: list index out of range"```. I was looking for a solution and I realised that it was because the kivy string (line 36) loaded the ```self.color``` as ```color = ListProperty([])```, even if it is modified later in the ```__init__``` of the ```MDSpinner()``` class.

I found three solutions:

1. The one you can see in this Pull Request, which basically obtains the ```self.theme_cls.primary_color``` directly from the ```ThemeManager()``` class (in case the user does not specify any ```primary_color```, it defaults to ```Blue```).

1. Specifying the color with ```color=[r, g, b, a]``` when calling the class (for example, ```MDSpinner(color=[1, 1, 1, 1])```, where is later replaced with ```self.theme_cls.primary_color```)

1. Changing the line 78 to ```color = ListProperty([r, g, b, a])``` (for example: ```ListProperty([1, 1, 1, 1])```), which will be later replaced with ```self.theme_cls.primary_color```.

By the way, the color can be modified through the attribute ```MDSpinner().color```. For example:
```
spinner = MDSpinner()
spinner.color = [1, 1, 1, 1]
self.add_widget(spinner)
```

I hope this help.

### Screenshots

MDSpinner

### DIFF
--- a/kivymd/uix/spinner.py
+++ b/kivymd/uix/spinner.py
@@ -21,6 +21,7 @@ from kivy.uix.widget import Widget
 from kivy.properties import NumericProperty, ListProperty, BooleanProperty
 from kivy.animation import Animation
 from kivymd.theming import ThemableBehavior
+from kivymd.theming import ThemeManager
 
 Builder.load_string(
     """
@@ -74,7 +75,7 @@ class MDSpinner(ThemableBehavior, Widget):
     defaults to True
     """
 
-    color = ListProperty([])
+    color = ListProperty(ThemeManager().primary_color)
     """:attr:`color` is a :class:`~kivy.properties.ListProperty` and
     defaults to 'self.theme_cls.primary_color'
     """


HeaTTheatR:
Then you need to edit this line too - https://github.com/HeaTTheatR/KivyMD/blob/94cd92103b69accd7698cd1f68bb566c14464bf2/kivymd/uix/spinner.py#L90

kivymd/uix/spinner.py
```diff
@@ -74,7 +75,7 @@ class MDSpinner(ThemableBehavior, Widget):
     defaults to True
     """
 
-    color = ListProperty([])
+    color = ListProperty(ThemeManager().primary_color)
```
ArtemSBulgakov, line 78: ThemeManager object will be created on each initialization of MDSpinner. It is not good

Franccisco: Oh, sorry. You're right. I think there's another approach:
We can change the inheritance of the MDSpinner to only ThemeManager and replace all the self.theme_cls to only self. This way, the line 78 can be kept as the original one (```color = ListProperty([])```) and the code works fine.

ghost (commented):
> Then you need to edit this line too -
> 
> https://github.com/HeaTTheatR/KivyMD/blob/94cd92103b69accd7698cd1f68bb566c14464bf2/kivymd/uix/spinner.py#L90



> Then you need to edit this line too -
> 
> https://github.com/HeaTTheatR/KivyMD/blob/94cd92103b69accd7698cd1f68bb566c14464bf2/kivymd/uix/spinner.py#L90

PR #113, opened 2019-10-03
Icon for menu

zworkb:
### Description of Changes

- Add an icon property to MDMenuItem.
- Use MDLabel instead of Label, so the font color is correctly set by KivyMD's palette

### Examples

Using of the `icon` property

```python
       menu_items = [
            dict(
                viewclass='MDMenuItem',
                icon='file-document',
                text=self.text_show_properties,
                callback=lambda *a:self.show_props()
            ),
            dict(
                viewclass='MDMenuItem',
                icon='share-variant',
                text=self.text_share,
                callback=lambda *a: self.on_share()
            )
        ]

        context_menu = MDDropdownMenu(
            items=menu_items,
            max_height=dp(260), width_mult=4)

        context_menu.open(self)
```

### DIFF
--- a/kivymd/icon_definitions.py
+++ b/kivymd/icon_definitions.py
@@ -22,6 +22,7 @@ LAST UPDATED: Version 4.3.95
 __all__ = ("md_icons",)
 
 md_icons = {
+    "null-icon": "",
     "ab-testing": "\U000F001C",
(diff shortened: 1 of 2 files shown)


ArtemSBulgakov:
@zworkb Thank you for pull request! This is very useful feature.

PR #117, opened 2019-10-13
Update MDTimePicker to require a callback function

Dirk-Sandberg:
MDTimePicker now requires a callback function when initialized. The callback function is called when the 'ok' button is pressed, and the time is automatically passed as an argument.


Dirk-Sandberg:
MDSwitch now initializes properly. Before the thumb would not be in the correct position when the widget initialized. This fixes that issue for when the MDSwitch is made with both active: True or active: False.

HeaTTheatR:
You should also change the `MDTimePicker` calls in the `Kitchen Sink` application.

Dirk-Sandberg:
Good idea, I didn't think about that. Done!

HeaTTheatR:
Better then do this:

```python
class MDTimePicker(
    ThemableBehavior, FloatLayout, ModalView, RectangularElevationBehavior
):

    """MDTimePicker.

    :Events:
        `on_get_time`
    """

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.register_event_type("on_get_time")

    def on_get_time(self):
        return self.time

    def close_ok(self):
        self.current_time = self.ids.time_picker.time
        self.time = self.current_time
        self.dispatch("on_get_time")
        self.dismiss()
```

PR #124, opened 2019-10-22
Add IDs to MDThemePicker

noembryo:
I'm trying to use the `MDThemePicker` with i18n (or if you like, l10n ;o) and I need access to the picker's text.
There are 2 ways to do that now:
* Create my own picker, copying the stuff from the original and adding the _("Texts") (I use gettext)
* Find the texts using something like `MDThemePicker.children[0].children[1].children[0]`... for evry one of it, and hoping that they will stay in the same place in the future.

The best way for me would be a third way.
Having IDs for every widget with text in the kv

`MDDatePicker` and `MDTimePicker` already have them

### DIFF
--- a/kivymd/uix/picker.py
+++ b/kivymd/uix/picker.py
@@ -575,11 +575,13 @@ Builder.load_string(
 
 
     MDFlatButton:
+        id: close_button
         pos: root.pos[0]+root.size[0]-self.width-dp(10), root.pos[1] + dp(10)
         text: "Close"
         on_release: root.dismiss()
 
     MDLabel:
+        id: title
         font_style: "H5"
         text: "Change theme"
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@noembryo Soon, `picker` will be completely redone, as it works very slowly on mobile devices. Do not rely too much on his API.

HeaTTheatR:
@noembryo This applies only to `ThemePicker`

noembryo:
OK. Just start working with it, but the only things that I need are the IDs, as the `app.theme_cls.accent_palette`, `app.theme_cls.primary_palette` and `app.theme_cls.theme_style` are still going to change with the new one, aren't they ;o)

PR #126, opened 2019-10-23
Add ID to ContentMDDialog

noembryo:
This ID is needed if we want to create a custom `ContentMDDialog`

### DIFF
--- a/kivymd/uix/dialog.py
+++ b/kivymd/uix/dialog.py
@@ -167,6 +167,7 @@ Builder.load_string(
                 source: f'{images_path}transparent.png'
 
         MDLabel:
+            id: dialog_text
             text: '\\n' + root.text + '\\n'
             size_hint_y: None
             height: self.texture_size[1]


HeaTTheatR:
It is not clear why this is necessary.

noembryo:
Here is a method that I use to create an "About the app" dialog
```python
    def on_about(self, *__):
        """ Creates a popup with the "About" page
        """
        color = get_hex_from_color(self.app.theme_cls.primary_light
                                   if self.theme == "Dark"
                                   else self.app.theme_cls.primary_dark)
        text = _("BlaBlaBla\n\n\n\n\n\n")
        text = text.format(APP_NAME, __version__, color, str(int(sp(28))))
        content = ContentMDDialog(title=_("About {}").format(APP_NAME), text=text,)
        content.ids.title.theme_text_color = "Primary"
        content.ids.dialog_text.theme_text_color = "Primary"
        content.ids.dialog_text.bind(on_ref_press=partial(webbrowser.open, APP_PAYPAL))

        self.popup = XMDDialog(size_hint=(.9, .7), text_button_ok=_("CLOSE"),
                               events_callback=lambda *__: __)
        self.popup.add_widget(content)
        self.popup.set_content(content)
        self.popup.bind(on_dismiss=self.about_popup_closed)
        self.popup.open()
```
The color of the text is black on the black theme.
If I don't have access to the text label there is no way to affect the color or to bind the link presses with an action.
If you want I can create a simple demo, but I think you can see that there is no way to access it.

HeaTTheatR:
This needs to be fixed not as you suggested ... And to make a `MDLabel` binding to changing the color theme of the application

noembryo:
Can you elaborate?
How can I bind the text's `refs` and change the color?

PR #150, opened 2019-12-24
Added icon_right re-color to the _on_focus method of MDTextFieldRound

RedGui:
Just a tiny PR with the below changes.

Changes:
- Changed the default on_focus behavior in `MDTextFieldRound` to also re-color the `icon_right` (to be coherent with existing `icon_left` behavior)
- Tied the `show_password` icon evaluate to the `password` Boolean property in the kitchen_sink demo (this also enables new functionality compared to before if `icon_right: 'account-box'`, meaning it has been declared as any icon other than `eye` or `eye-off` initially, this would then correctly change the icon from its initial icon to either `eye` or `eye-off` and then its counterpart, but allows for a different icon to exist in the input field's initial state)

### DIFF
--- a/demos/kitchen_sink/libs/baseclass/textfields.py
+++ b/demos/kitchen_sink/libs/baseclass/textfields.py
@@ -16,4 +16,4 @@ class KitchenSinkTextFields(Screen):
         # and set icon of right button.
         field.password = not field.password
         field.focus = True
-        button.icon = "eye" if button.icon == "eye-off" else "eye-off"
+        button.icon = ["eye", "eye-off"][field.password]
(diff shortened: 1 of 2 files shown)


RedGui:
This is still valid, resolved the conflicts with latest changes.

RedGui:
@HeaTTheatR GitHub failed to get the build status from Travis. The run completed.

RedGui:
@HeaTTheatR Could you add a comment as to why you chose not to merge this in?

It seems a bit spotty if only one of the two icons is getting recolored on focus, despite the rounded text field having two possible icons.

PR #165, opened 2020-01-13
Enable/disable past/future dates

shashi278:
Many-a-times we want users to restrict from being able to select past and/or future dates and thus I've added two optional parameters in the constructor function of `MDDatePicker` viz. `show_past_date` (defaults to `True`) & `show_future_date` (defaults to `True`).

I've also checked the updated code with the demo app and its working perfect.

### DIFF
--- a/kivymd/uix/picker.py
+++ b/kivymd/uix/picker.py
@@ -327,8 +327,8 @@ class MDDatePicker(
         month=None,
         day=None,
         firstweekday=0,
-        show_past_date=True,
-        show_future_date=True,
+        min_date=None,
+        max_date=None,
         **kwargs,
     ):
         self.callback = callback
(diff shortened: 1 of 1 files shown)


ArtemSBulgakov:
Great idea! I think that it is very useful thing. But it's better to have such properties as `min_date` and `max_date`. If you want to help us realize this, please rebase your branch and make new commits. Do not forget to format your code using Black utility. See [CONTRIBUTING](https://github.com/HeaTTheatR/KivyMD/blob/master/.github/CONTRIBUTING.md#i-have-a-changes).

shashi278:
Thanks! Actually, `min_date` and `max_date` suits better in this context and I can think of them as being `datetime` object or maybe `None` if no restriction is to be imposed. I'll update the code accordingly.

shashi278:
Can we have a reviewer for this? @Mountaino

ArtemSBulgakov:
Works fine

PR #192, opened 2020-02-29
Allowing to change the color of the bottom toolbar

Nibodhika:
### Description of Changes

* Fixing an issue with on_md_bg_color, the comparison was made with type instead of self.type

* Allowing to change the color of a Toolbar that has the bottom type. Originally it was hardcoded to primary_color, I added an internal property called _bottom_md_bg_color which is set instead of the md_bg_color when the type is bottom.

* Fixing the identation on the bar always being drawn in white, switching to use the actual bg color

### DIFF
--- a/kivymd/uix/toolbar.py
+++ b/kivymd/uix/toolbar.py
@@ -298,7 +298,7 @@ Builder.load_string(
                 if root.type == "bottom" else (0, 0)
             radius: (root.round, 0, 0, 0) if root.mode == "center" else (0, root.round, 0, 0)
         Color:
-            rgba: 1, 1, 1, 1
+            rgba: root.theme_cls.bg_normal
         Ellipse:
             pos:
                 (self.center[0] - root.action_button.width / 2 - dp(6), self.center[1] - root._shift * 2) \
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@Nibodhika The `_bottom_md_bg_color` parameter is not required. In addition, we must adhere to the general specification for the dos string format, as in the Kivy library.

Nibodhika:
Hi, thanks for the feedback. Let me see if I understood correctly, what I would need to do is:

* Remove the added lines 396-401

* ~~Change the variable name from `_bottom_md_bg_color` to something else (I couldn't find the documentation for the dos string format, so I'm not sure how to call the variable, could you suggest a name and/or link that documentation to me please?)~~ Just realized you meant docs string, see children, this is why you don't reply before drinking a cup of coffee first. In any case what was wrong with the docstring?

* Declare it in the `__init__` instead to ensure is populated.

HeaTTheatR:
The `_bottom_md_bg_color` parameter is not needed!

Nibodhika:
Then how would you suggest the issue with md_bg_color requiring to be different from the colour of the background of the bar?

PR #226, opened 2020-03-25
Allow dropdownitem to be disabled

adamskrz:
### Description of Changes
* Adds 'disabled' attribute
* Stops menu pop-up if disabled is True
* Allow set_item to be used externally without opening the menu (eg when disabled)


HeaTTheatR:
I do not see any reason in supporting this module, because soon it will be completely rewritten in accordance with the requirements of material design.

PR #241, opened 2020-04-07
Made it to directly animate the object itself

shashi278:
Now it directly animates the object itself rather than depending upon its .kv file


shashi278:
Test the modified implementation and let me know if it works as required

PR #255, opened 2020-04-19
Remove hardcoding of ext

99hats:
Left-over test code?

### Description of Changes
Looks like self.ext is hardcoded limiting the utility of the widget.


HeaTTheatR:
See where `self.ext` is used

PR #294, opened 2020-05-08
Update text color of labels

idevo89:
Text color of the labels of MDFloatingActionButtonSpeedDial gets updated when changing value


HeaTTheatR:
@idevo89 What do your changes bring to the code?

idevo89:
image

Change background of theme to dark and you'll notice the text color of the labels is not the same like the icon color because the label text color will follow the inverted background color.

HeaTTheatR:
@idevo89 Thanks. I will test these changes later ...

ArtemSBulgakov:
@idevo89 Now it works as expected, thanks

PR #298, opened 2020-05-11
fitimage: add texture support

wasimafser:
some people would use texture only for images and this would be helpful for them too.

### Description of Changes
* ...
* ...

### Screenshots

Add images to explain us your changes. Paste urls here.

Remove this section if no images here

### DIFF
--- a/kivymd/utils/fitimage.py
+++ b/kivymd/utils/fitimage.py
@@ -36,9 +36,11 @@ from kivy.uix.widget import Widget
 class FitImage(BoxLayout):
     source = StringProperty()
     texture = ObjectProperty()
+    container = None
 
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@wasimafser Thanks. I will test these changes later ...

wasimafser:
Sure, thank you ! just a small change though !

UPDATE : Done

ArtemSBulgakov:
@wasimafser Thank you for your PR. I decided to implement it by self because it is faster.
To prevent setting two variables (`source` and `texture`) at one time I decided to use only one `source`. Also I added binding to dynamically set image (as suggested in #348).

PR #319, opened 2020-05-21
add navigationlayout scrim using clock

sanderland:
# Description of Changes
* Add a delay to adding the scrim, avoiding issues with incorrect size

### Screenshots
Before:
image

After:
image

### DIFF
--- a/kivymd/uix/navigationdrawer.py
+++ b/kivymd/uix/navigationdrawer.py
@@ -275,6 +275,7 @@ __all__ = ("NavigationLayout", "MDNavigationDrawer")
 
 from kivy.core.window import Window
 from kivy.logger import Logger
+from kivy.clock import Clock
 from kivy.animation import Animation, AnimationTransition
 from kivy.graphics.context_instructions import Color
 from kivy.graphics.vertex_instructions import Rectangle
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
There is no description of the changes!

sanderland:
> 
> 
> There is no description of the changes!

Yes there is!

HeaTTheatR:
@sanderland I do not understand what you did. I don’t understand why you did it. You have not provided any information about your changes.

sanderland:
> @sanderland I do not understand what you did. I don’t understand why you did it. You have not provided any information about your changes.

I provided a description and two screenshots, for a one line change.
Perhaps the first image was not clear, here the problem is circled.

image

Kivy has a very confusing ordering of setting properties, and in this case seems to initialize size to something weird, and then change size without firing the event. The clock delay usually prevents this.

HeaTTheatR:
@sanderland There is no such problem!

sanderland:
> 
> 
> @sanderland There is no such problem!

Then what is causing the difference in the screenshots?

HeaTTheatR:
@sanderland Your code!

sanderland:
> @sanderland Your code!

The root cause appeared to be using a Widget as the root level object instead of a Screen.

HeaTTheatR:
@sanderland https://kivymd.readthedocs.io/en/latest/components/navigation-drawer/index.html

sanderland:
> 
> 
> @sanderland https://kivymd.readthedocs.io/en/latest/components/navigation-drawer/index.html

Yes, the word 'Root' there makes it seems it is more flexible, changing it to screen as in the example may avoid this.

PR #337, opened 2020-06-01
MDLabel set default theme_text_color to "Primary"

allasca:
### Description of Changes
Set default MDLabel theme_text_color to "Primary"

### DIFF
--- a/kivymd/uix/label.py
+++ b/kivymd/uix/label.py
@@ -278,7 +278,7 @@ class MDLabel(ThemableBehavior, Label):
     """Text of the label."""
 
     theme_text_color = OptionProperty(
-        None,
+        "Primary",
         allownone=True,
         options=[
             "Primary",


HeaTTheatR (changes requested):
Have you tested these changes in the `KitchenSink` app?

allasca:
@HeaTTheatR sorry if I wrong or this is not needed. but yes, result was like this

```python
from kivymd.app import MDApp
from kivy.lang import Builder
from label import MDLabelku

KV = '''
BoxLayout:
    MDLabelku:
        text: "test 1234567"
    Button:
        text: "change background"
        on_press: app.change()
'''

class Label(MDLabelku):
    pass

class MyApp(MDApp):
    def build(self):

    def change(self):
        self.theme_cls.theme_style = "Dark"


MyApp().run()
```

PR #348, opened 2020-06-09
Binding Image.source to FitImage.source

BernardinD:
### Description of Changes
Added a binding to between the nested Image and the FitImage object so that the image can be dynamically updated


HeaTTheatR:
@BernardinD This does not work. In addition, your code is not formatted correctly - https://github.com/psf/black

PR #357, opened 2020-06-17
update filemanage.py

PunitSolanki1999:
self.ext is not taking user given extension because default initilization of self.ext in def __init__(), Please check it out.
Thank you.


PunitSolanki1999:
I hope my help will be userful to you.

HeaTTheatR:
This will only work for the `previous = False` mode.

PR #361, opened 2020-06-21
Bug fix for progressloader.py

sanderland:
Small bugfix for default value having the wrong number of parameters.


HeaTTheatR:
@sanderland Thank you for your pull request! But this module will be removed in the next version.

PR #409, opened 2020-07-19
Fixed Menu height bug occuring after window resize

kevin-schumann:
### Description of Changes
Replacing bugfix in commit 6d9b448  with a new fix.
Proposed change fixes menu height growth both when
reopening the menu and resizing the window.

Former bugfix only resolves the former issue.


HeaTTheatR:
There is no fundamental difference between your first pool and your second. Your first pool solves the problem.

PR #447, opened 2020-08-09
Bugfix MDTogglebutton behavior

podraco:
### Description of Changes
The togglebutton module had a bug in wich the buttons created with this
behavior didn't changed even if they where in a group, the colors got
stuck and the overall selection of them was flawn.

This fix uses a fbind to the `state` property, so everytime is changed,
the funciton will be called without being overwritten by the on_release
of the new modules. and also updates the selection if it's state is
changed from outside the class.

`__is_filled` is a boolean memory that turns to true if the custom
object has background filling, this allows to change the font color of
the custom widgets when they're pressed down, the color chagnes to white
and turnd back to the theme_cls.primary_color once it's state is normal.

for the moment this fixes are strict/rigid, but in future revisions it
might become more flexible around customized themes.

MDToggleButton.theme_cls has been added. to save some flops everytime
the MDApp.get_running_app is called to retrieve the app.theme_cls.

At the end of the module was added an ifmain codeblock to allow future
developers to test out the changes faster and in a safe way.

This TEST app uses the instance of the kivymd installed in the
developers pc but the module tested (in this case MDtogglebutton) is the
one in the file.

the module has been documented to help future developres know what's
happening in every line that's not intituive enough.

Some inline comments has been added past the char 80 to allow a better
understanding of the code documented.


HeaTTheatR:
@podraco 
Use `pre-commit run --all-files` - https://github.com/kivymd/KivyMD#setup-environment
No need to use block `if __name__ == __main__` - we have the documentation and that should be enough.

podraco:
I'll let the `if __name__ == __main__` code block since it lets the developers see if our fixes really works and doesn't mess up something else. Also, allows us and new programmers to interact with the widget directly. 

This is the same practice the kivy team uses with their files. You can easily call `python -m kivy.some.module` and it will show you a sample app where you can test the widget.

HeaTTheatR:
1. Base class of buttons - already has a property `theme_cls` Therefore, the following code is not needed:

```python
class MDToggleButton(ToggleButtonBehavior):
    # THeming
    theme_cls = ObjectProperty()

    ...

    def __init__(self, **kwargs):
        ...

        # Verify: Do The app class has theme_cls?
        app = MDApp.get_running_app()  # This saves some flops
        if not hasattr(app, "theme_cls"):
            raise ValueError(  # If the app doesn't has the theme_cls, it will raise a ValueError exception with message
                "KivyMD: App object must be inherited from "
                "`kivymd.app.MDApp`. See "
                "https://github.com/kivymd/KivyMD/blob/master/README.md#api-breaking-changes"
            )

```

2. Why don't you adhere to the `PEP8` standards and limit the line of code to 79 characters?

3. Delete block `if __name__ == __main__` - we do not follow this practice.

podraco:
Ok, so I removed the Theme_cls checks from the ToggleButtonBehavior class, indeed the other buttons gave the property and i didn't check that before, Thanks.

Responding to:
>Why don't you adhere to the PEP8 standards and limit the line of code to 79 characters?

I've been taught to use the 80 char limiter to write pseudo code when I'm testing it gives the code more legibility. Since i'ts discarded when precompiled so It doesn't affect performance. But I think I fixed it to be pep8 compliant.

>Delete block if __name__ == __main__ - we do not follow this practice.

Why don't we follow this practice? It literarily avoids creating more bugs when we're applying some patch.
Otherwise, how are we supposed to catch a bug derived from a modification recently added?

If the documentation code where enough, then why was there a bug that big?

HeaTTheatR:
> Ok, so I removed the Theme_cls checks from the ToggleButtonBehavior class, indeed the other buttons gave the property and i didn't check that before, Thanks.

```python
157.        else:
158.             self.theme_cls = MDApp.get_running_app().theme_cls
```
This is not needed code. Base class of buttons - already has a property `theme_cls`

> Why don't we follow this practice? It literarily avoids creating more bugs when we're applying some patch.
Otherwise, how are we supposed to catch a bug derived from a modification recently added?

Because initially we did not have this practice. And if you propose to do this, then this should be done in each module of the library, and not only in one. Therefore, either we use this practice everywhere or we do not use it anywhere.

podraco:
```python
                [...]
                else:
                    self.theme_cls = MDApp.get_running_app().theme_cls
                [...]
```
>This is not needed code. Base class of buttons - already has a property theme_cls

Sorry, forgot about that one block

>Because initially we did not have this practice. And if you propose to do this, then this should be done in each module of the library, and not only in one. Therefore, either we use this practice everywhere or we do not use it anywhere.

Ideally that would be the best practice. But for now, I'll remove it, until I have more free time, in the meanwhile, i'll let it as part of the commit history in case we want to add the test code in the future.

As part of the practice, ill be doing testapp's Codeblocks for every file i check and modify to fix or add new capabilities, and do the same. Checking all works and then removing the testapp codeblock to keep it in the commits' history.

PR #450, opened 2020-08-10
Fixed table crash when going back in forth the pagination

kevin-schumann:
### Description of Changes
Fixed index out of bounds when using pagination for the following condition:

`number of rows % rows per page != 0`

### Reproduce

1.  Have table with 9 rows
2.  Set rows per page 5
3.  Go to the last page
4.  Go the the first page
5.  do 3. again

### DIFF
--- a/kivymd/uix/datatables.py
+++ b/kivymd/uix/datatables.py
@@ -567,7 +567,7 @@ class TableData(RecycleView):
                     self._row_data_parts[self._rows_number]
                 )
                 self._to_value = self._to_value - len(
-                    self._row_data_parts[self._rows_number]
+                    self._row_data_parts[self._rows_number + 1]
                 )
             if direction == "increment":
                 self._current_value = 1


HeaTTheatR:
@kevin-schumann Thanks for helping to improve the library!
Could you please attach code examples when the error you fixed occurs.

kevin-schumann:
****

### Minimal code:

**Put the window in fullscreen as there is another bug (scrollview) that causes the programm to crash.**

```python
from kivy.metrics import dp

from kivymd.app import MDApp
from kivymd.uix.datatables import MDDataTable


class Example(MDApp):
    def build(self):
        self.data_tables = MDDataTable(
            size_hint=(0.9, 0.6),
            use_pagination=True,
            check=True,
            column_data=[
                ("No.", dp(30)),
                ("Column 1", dp(30)),
                ("Column 2", dp(30)),
                ("Column 3", dp(30)),
                ("Column 4", dp(30)),
                ("Column 5", dp(30)),
            ],
            row_data=[
                (f"{i + 1}", "2.23", "3.65", "44.1", "0.45", "62.5")
                for i in range(7)
            ],
        )

    def on_start(self):
        self.data_tables.open()


Example().run()
```

### Instructions:

1.  Press pagination forward till it gets disabled
2.  Press pagination backward till it gets disabled
3.  Press pagination forward till it gets disabled

PR #457, opened 2020-08-14
Allows use of md_icons when using icon property of MDExpansionPanel

p0lygun:


ArtemSBulgakov:
@p0lygun Please remove commit with edits to Readme and force push to your fork

HeaTTheatR:
@p0lygun If the `self.icon` parameter does not exist you must set it a default value.

Issue #345, opened 2020-06-04
MDRaisedButton md_bg_color attribute won't work as expected


mozesa:
### Description of the Bug

You cannot instantiate a new `MDRaisedButton` in python with altered color even though `md_bg_color` parameter is provided with a new color.

### **NOK** as parameter
```python
from kivy.uix.screenmanager import Screen
from kivymd.app import MDApp
from kivymd.uix.button import MDRaisedButton


class App(MDApp):
    def build(self) -> None:
        screen = Screen()
        raised_button = MDRaisedButton(
            text="Hello",
            md_bg_color=[1, 0, 0, 1],
            pos_hint={"center_x": 0.5, "center_y": 0.5},
        )
        screen.add_widget(raised_button)
        self.root = screen


app = App()
app.run()
```
![image](https://user-images.githubusercontent.com/49973426/83768034-9ab71000-a67e-11ea-91e8-15677a05d7da.png)

---

### **OK** as attribute set
```python
from kivy.uix.screenmanager import Screen
from kivymd.app import MDApp
from kivymd.uix.button import MDRaisedButton


class App(MDApp):
    def build(self) -> None:
        screen = Screen()
        raised_button = MDRaisedButton(
            text="Hello", pos_hint={"center_x": 0.5, "center_y": 0.5},
        )
        raised_button.md_bg_color = [1, 0, 0, 1]
        screen.add_widget(raised_button)
        self.root = screen


app = App()
app.run()
```
![image](https://user-images.githubusercontent.com/49973426/83767860-6a6f7180-a67e-11ea-8220-e6fe752715d6.png)

### Versions

* OS: Windows 10
* Python: 3.8.3
* Kivy: v2.0.0rc2
* KivyMD: v0.104.1 (master)

HeaTTheatR:
@mozesa So far it will work that way.

podraco:
Does this mean the widget doesn't check if there's already a set for md_bg_color before the instance?

podraco:
i believe that this might solve a problem found in the test app of #447 that I think is derived from this strange behavior

PR #480, opened 2020-08-21
WIP Improve MDRaisedButton md_bg_color behavior

podraco:
### Description of Changes
fix to solve MDRaisedButton md_bg_color attribute won't work as expected #345 

#### Improve MDRaised Button.

The MDRaised button now allows color theming.
Allowing the Custom color and Primary color 

The theme_button_color has a similar Behavior to 
MDLabel.theme_text_color

Documentation has been added to help developers understand the changes.

A test app has been added at the end of the file to test the script, it 
will be removed once the push merges is approved.
##### Additions
1. Adds a `MDRaisedButton.md_bg_color` ListProperty (is no longer inherited from the base class)
1. Adds a `MDRaisedButton.theme_button_color` OptionProperty
1. Adds behavior to the new properties 
1. Edits `MDRaisedButton.__init__` to set automatically the theme_button_color property
1. Adds Documentation to MDRaisedButton

#### Note:
This pull request contains a test app that will be removed once the changes are approved

### DIFF
--- a/kivymd/uix/button.py
+++ b/kivymd/uix/button.py
@@ -707,7 +707,7 @@ Builder.load_string(
 
 
 <MDRaisedButton>
-    md_bg_color: root.theme_cls.primary_color
+    md_bg_color: root.md_bg_color
     theme_text_color: 'Custom'
     text_color: root.specific_text_color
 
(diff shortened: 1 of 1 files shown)


HeaTTheatR (commented):
Thank you for helping to improve the `KivyMD` library.
Your changes are based on an existing problem. And they don't fix this problem, but ignore it. Changing the `md_bg_color` parameter until recently worked fine in `Python` code when instantiating a button. And our task is to figure out why this does not work now.

podraco:
ho, it's quite easy, the parameter is a herence from the base button behavior.
since it's defined in kv lang the new instances already has a md_bg_color thats [1,1,1,0]
to fix this the md_bg_color must be [] before the instance, but this might mess up with other widgets that work differently

i velieved the problem comes from line 710
```
-    md_bg_color: root.theme_cls.primary_color
+    md_bg_color: root.md_bg_color
```

HeaTTheatR:
@podraco Do you not understand the essence of the problem?

podraco:
Well now I have a doubt, could you explain to me the essence of the problem? 
Maybe I'm looking to it in the wrong way...

HeaTTheatR:
@podraco You haven't fixed the problem. You ignored her. If you fixed the existing problem, your changes would not be needed. Plus, you apply the changes to one specific button class.

kivymd/uix/button.py
```diff
@@ -707,7 +707,7 @@ def build(self):
 
 
 <MDRaisedButton>
-    md_bg_color: root.theme_cls.primary_color
+    md_bg_color: root.md_bg_color
```
podraco, line 710: This is the main problem, 
the kv code already gave a md_bg_color, 
this launch the on_md_bg_color
and obstructs the event of a new instance.

kivymd/uix/button.py
```diff
@@ -1101,7 +1101,86 @@ class MDRaisedButton(
...
+        super(MDRaisedButton, self).__init__(**kwargs)
+        if self.theme_button_color == "None":
+            if not self.md_bg_color:
+                self.theme_button_color = "Primary"
+                self.md_bg_color = self.theme_cls.primary_color
+            else:
+                self.theme_button_color = "Custom"
+        elif self.theme_button_color == "Primary":
+            self.md_bg_color = self.theme_cls.primary_color
+            # self._current_button_color = self.theme_cls.primary_color
```
podraco, line 1168: This code detects if the argument md_bg_color is given before the instance, before the change calls the on_ event and sets a new behavior to allow custom color buttons.

podraco (commented):
its the only class i found it's working wrong, haven't tested with other buttons class
but you are right, i must check the other buttons before trying to make a pull request.

kivymd/uix/button.py
```diff
@@ -1101,7 +1101,86 @@ class MDRaisedButton(
...
+        # This allows to use set a cusotm color and enable or disable at will.
+        if self.theme_button_color == "Primary":
+            self._current_button_color = self.theme_cls.primary_color
+        elif self.theme_button_color == "Custom":
+            self._current_button_color = self.md_bg_color
+
+    def update_md_bg_color(self, instance, value):
+        """Called when the application color palette changes."""
+        if self.theme_button_color == "Primary":
+            self._current_button_color = self.theme_cls._get_primary_color()
```
podraco, line 1183: this part is a patch to another bug, i must say
this one updates all buttons since they dont have a way to know when to avoid the event of the theme change

kivymd/uix/button.py
```diff
@@ -1840,3 +1919,63 @@ def _get_count_widget(self, instance):
...
+        def change_theme_color2(self, *dt):
+            self.theme_cls.primary_palette = "Red"
+
+        #
+        def change_theme_color3(self, *dt):
+            self.theme_cls.primary_palette = "Blue"
+            self.raised_button1.theme_button_color = "Primary"
+
+    app = App()
+    app.run()
```
podraco, line 1981: This code will be removed in future revitions

podraco (commented):
update_md_bg_color is called even if the button is not custom color

PR #508, opened 2020-09-06
Fixed Nav Layout not shrinking in Modal

dwmoffatt:
### Description of Changes
checking the NavLayout tDrawer type then set manager position and width accordingly


dwmoffatt:
Is there anything else that I need to do from my end for this? Is how I have done it acceptable?

PR #520, opened 2020-09-12
change textfield animation that turns hint_text black and other modifications

### LINKED ISSUE #448: MDTextField class not working as spected
### Description of the Bug

This is a list of errors i encountered while working with the MDTextfield widget class.

#### Open

2. The filling is drawing after all the textures. if the filling has an Alpha of 1, the whole widget will be turned the color, hiding everything inside the widget.
- This could be solved by either drawing the filling in the canvas.before
This might cause the text to be illegible over the filling, since it's not taking any color from it, it might be a good idea to drop the alpha channel of the text to 0.9 to allow the text to merge and gain some strength from the filling, as it happens with the filling example with the filling color of `[0, 0, 0, 0.4]`
- Or parsing the filling color as an RGB format and adding a default Alpha of 0.5.

3. The MDTextField doesn't change the text to black when unfocused in any iteration of the properties 
This will mislead the end user to believe he's changing a field that is not on focus. Or in the case that the UI has multiple fields, loss track of what was the end-user doing.

4. The rectangle of MDTextField when `mode: "rectangle"` is not changing color when unfocused or on_focus.
As the point before, this will mislead the end-user  to believe that they had selected a text field not on focus.

5. The filled MDTextField and `mode: "rectangle"` draws the instruction outside the widget, it requires an extra padding of dp(16) around it to show properly.

### update 2020-08-10-6:33 AM GTM-6

When the MDTextField canvas.before is modified (added a new instruction), the last added color is used to draw the text. This might lead to the bug. Since it might be caused because the text color is not settled.

#### Fixed
1. The `hint_text `object doesn't work properly when `on_error `status is active, if the MDTextfield had text inside once, the color of the `hint_text `will turn black instead of red. but before the MDTextfield had any text it worked properly.
This behavior is consistent with every error status, independent if it's launched by `requider:True`, `max_text_length`or manually set to error.
It seems that the red text is being drawn below the black text as seen in the pictures below, might have something to do with the texture of the label.
FIXED IN #520 

### Code and Logs

```python
from kivy.app import App
from kivy.lang import Builder

from kivymd.app import MDApp

KV="""
BoxLayout:
    padding:dp(8)
    MDGridLayout:
        cols:2
        spacing:dp(16)
        MyTextField:
            hint_text: "No helper text"
        MyTextField:
            hint_text: "Helper text on focus"
            helper_text: "This will disappear when you click off"
            helper_text_mode: "on_focus"
        MyTextField:
            hint_text: "Persistent helper text"
            helper_text: "Text is always here"
            helper_text_mode: "persistent"
        MyTextField:
            id: text_field_error
            hint_text: "Helper text on error (press 'Enter')"
            helper_text: "There will always be a mistake"
            helper_text_mode: "on_error"
            pos_hint: {"center_y": .5}
        #
        MyTextField:
            hint_text: "required = True"
            required: True
            helper_text_mode: "on_error"
            helper_text: "Enter text"
        MyTextField:
            hint_text: "Max text length = 5"
            max_text_length: 5
        MyTextField:
            multiline: True
            hint_text: "Multi-line text"
        MyTextField:
            hint_text: "color_mode = 'accent'"
            color_mode: 'accent'
        MyTextField:
            hint_text: "color_mode = 'custom'"
            color_mode: 'custom'
            helper_text_mode: "on_focus"
            helper_text: "Color is defined by 'line_color_focus' property"
            line_color_focus: 1, 0, 1, 1
        MyTextField:
            hint_text: "Line color normal"
            line_color_normal: app.theme_cls.accent_color
        MyTextField:
            hint_text: "Rectangle mode"
            mode: "rectangle"
        MyTextField:
            hint_text: "Fill mode"
            mode: "fill"
            fill_color: 0, 0, 0, .4
"""

class MyTextField(MDTextField):
    pass


class Test(MDApp):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def build(self):
        self.screen.ids.text_field_error.bind(
            on_text_validate=self.set_error_message,
            on_focus=self.set_error_message,
        )
        return self.screen

    def set_error_message(self, instance_textfield):
        self.screen.ids.text_field_error.error = True

Test().run()


if __name__ == '__main__':
    MainApp().run()
```

### Screenshots

in this images you can see that the black text has somewhat a red dye around the text, 
the circle is the zoomed part of the cursor, taken with share X.
In this image the error state is launched by `required: True`.
ShareX_MfusgeyTWM
In this image the error state is launched by `max_text_length: 5`.
ShareX_8kr9aCg4it

In this image the filling color is `[0, 0, 0, 1]` text `"asd"`
ShareX_CZx6Rgrsf5
In this image the filling color is `[0, 0, 0, 0.4]` text `"asd"`
ShareX_QIhZWWrDYe

In this image you can see that every TextField is not on focus, yet the color of the text is different from a normal color.
python_gZp7RhB2q9

In this image, you can see that the box is on primary color instead of a semi transparent shade of gray (ideally `[0,0,0,0.2]`)
python_tTNl3tReXH

In this image the widget is on focus
python_AqJjYbRCRA

In this image the widget is not on focus, and the box is not turning gray
python_ZSbchlpX85

Without enough padding, the Graphic instructions overlap everything else around them.
ShareX_Q7RvMcg038

### Versions

* OS:  Windows 10 rev 1909
* Python:  3.7
* Kivy: 1.11.1
* KivyMD:  master branch

CoelhoNL:
### Description of Changes

1. Fixed #448 issue 1.
 The animation on textfield textfield.py line 942
`                Animation(duration=0.2, color=self.line_color_focus).start(
                    self._hint_lbl
                )`
should change hint_text's color, but in reality, it adds self.line_color_focus on hint text's previous color, acting like a filter. This behavior makes hint text's color turning into a darker color.
I've changed the animation to a propper way that really changes the colors:
`Animation(
                    duration=0.2,
                    _current_hint_text_color=self.line_color_focus
                ).start(self)`

2. Changed the way that textfield acts when detects an error. 
When you get focus on a textfield and it's on error, the usual behavior is to turn hint_text's color to `line_color_focus`, overriding error_color. I've changed this adding an if statement on textfield.py line 909 to preserve error_color until widget's error is fixed.

### Screenshots

Before: 
image

After:
image

### DIFF
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -906,14 +906,15 @@ class MDTextField(ThemableBehavior, TextInput):
             )
             _fill_color = self.fill_color
             _fill_color[3] = self.fill_color[3] - 0.1
-            Animation(
-                _line_blank_space_right_point=self._line_blank_space_right_point,
-                _line_blank_space_left_point=self._hint_lbl.x - dp(5),
-                _current_hint_text_color=self.line_color_focus,
-                fill_color=_fill_color,
-                duration=0.2,
-                t="out_quad",
-            ).start(self)
+            if not self._get_has_error():
+                Animation(
+                    _line_blank_space_right_point=self._line_blank_space_right_point,
+                    _line_blank_space_left_point=self._hint_lbl.x - dp(5),
+                    _current_hint_text_color=self.line_color_focus,
+                    fill_color=_fill_color,
+                    duration=0.2,
+                    t="out_quad",
+                ).start(self)
             self.has_had_text = True
             Animation.cancel_all(
                 self, "_line_width", "_hint_y", "_hint_lbl_font_size"
@@ -939,9 +940,10 @@ class MDTextField(ThemableBehavior, TextInput):
                     self._anim_current_error_color(disabled_hint_text_color)
             else:
                 self._anim_current_right_lbl_color(disabled_hint_text_color)
-                Animation(duration=0.2, color=self.line_color_focus).start(
-                    self._hint_lbl
-                )
+                Animation(
+                    duration=0.2,
+                    _current_hint_text_color=self.line_color_focus
+                ).start(self)
                 if self.helper_text_mode == "on_error":
                     self._anim_current_error_color((0, 0, 0, 0))
                 if self.helper_text_mode in ("persistent", "on_focus"):
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -940,10 +940,9 @@ class MDTextField(ThemableBehavior, TextInput):
                     self._anim_current_error_color(disabled_hint_text_color)
             else:
                 self._anim_current_right_lbl_color(disabled_hint_text_color)
-                Animation(
-                    duration=0.2,
-                    _current_hint_text_color=self.line_color_focus
-                ).start(self)
+                Animation(duration=0.2, _current_hint_text_color=self.line_color_focus).start(
+                    self
+                )
                 if self.helper_text_mode == "on_error":
                     self._anim_current_error_color((0, 0, 0, 0))
                 if self.helper_text_mode in ("persistent", "on_focus"):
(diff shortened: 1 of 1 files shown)


podraco:
This change looks elegant and simple, good job!
I haven't had the opportunity to test it out yet

podraco:
image
i'm getting a dark hue below the text line, is that normal or is it only me?

PR #537, opened 2020-09-22
Nav layout closing incorrectly with swipe

### LINKED ISSUE #536: Wrong animation when Navigation Drawer being set to the right
when setting the drawer to open from the right side of the screen:
        anchor: "right"
the closing and opening swipe gestures remains as if the drawer is still on the left side of the screen.

ezgif com-video-to-gif

dwmoffatt:
### Description of Changes
Detect what side you are anchored to and the do the correct math to set open_drawer

Fixes #536

### DIFF
--- a/kivymd/uix/navigationdrawer.py
+++ b/kivymd/uix/navigationdrawer.py
@@ -696,14 +696,15 @@ class MDNavigationDrawer(MDCard):
                 self.status = "closing_with_swipe"
 
         if self.status in ("opening_with_swipe", "closing_with_swipe"):
-            if self.anchor == "left":
-                self.open_progress = max(
-                    min(self.open_progress + touch.dx / self.width, 1), 0
-                )
-            else:
-                self.open_progress = max(
-                    min(self.open_progress - touch.dx / self.width, 1), 0
-                )
+            self.open_progress = max(
+                min(
+                    self.open_progress
+                    + (touch.dx if self.anchor == "left" else -touch.dx)
+                    / self.width,
+                    1,
+                ),
+                0,
+            )
             return True
         return super().on_touch_move(touch)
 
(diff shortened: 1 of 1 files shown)


podraco:
i'll have to ask you to post a test code, so we will be able to test it out correctly.
and elaborate a little more this pull request, since issues and pull request are indexed separately.

kivymd/uix/navigationdrawer.py
```diff
@@ -696,9 +696,14 @@ def on_touch_move(self, touch):
...
-                min(self.open_progress + touch.dx / self.width, 1), 0
-            )
+            if self.anchor == "left":
+                self.open_progress = max(
+                    min(self.open_progress + touch.dx / self.width, 1), 0
+                )
+            else:
+                self.open_progress = max(
+                    min(self.open_progress - touch.dx / self.width, 1), 0
+                )
```
podraco: The fix works but,
it's more efficient it you only change the value that's affecting the code, in this case
instead of doing
```py
 if self.anchor == "left":
                self.open_progress = max(
                    min(self.open_progress + touch.dx / self.width, 1), 0
                )
            else:
                self.open_progress = max(
                    min(self.open_progress - touch.dx / self.width, 1), 0
                )
```
i would say the code should be:
```py
self.open_progress = max(
                min(
                    self.open_progress
                    + (touch.dx if self.anchor == "left" else -touch.dx)
                    / self.width,
                    1,
                ),
                0,
            )
```

podraco (commented):
As a side note, the interpreter works faster with inline if else clauses, since the whole if else blocks takes a little more analysis time.

but it's up to you.

kivymd/uix/navigationdrawer.py
dwmoffatt: Next time ill post some test code for the solution. I didn't know this "interpreter works faster with inline if else clauses, since the whole if else blocks takes a little more analysis time." I will for sure make the improvement mentioned abouve and re-push those changes.

podraco:
Those are some strange behavior, that are not intended, but came out as a product of the implementation.
I'm not sure if this trick keeps on pypy or other interpreter

dwmoffatt:
> Those are some strange behavior, that are not intended, but came out as a product of the implementation.
> I'm not sure if this trick keeps on pypy or other interpreter

not sure what you mean by your above statements @podraco

PR #538, opened 2020-09-22
menu with MDToolbar (CustomToolbar) improvement

kulothunganug:
### Description
when use [custom mdtoolbar](https://kivymd.readthedocs.io/en/latest/components/menu/index.html#menu-with-mdtoolbar) for dropdownmenu the labels and icon colours are in black always (see the below screenshots)
but in original kivymd's mdtoolbar changes textcolor of labels (and icons) according to theme_cls.primary_palette

### Screenshots
in all theme (primary_palette) mdlabels and mdicons in black in color
bad1        bad2

according to the current theme (primary_palatte) mdlabels and mdicons changes colors to specific_text_color 
gud1        gud2

### DIFF
--- a/kivymd/uix/menu.py
+++ b/kivymd/uix/menu.py
@@ -459,11 +459,14 @@ Menu with MDToolbar
         height: self.theme_cls.standard_increment
         padding: "5dp"
         spacing: "12dp"
+        md_bg_color: app.theme_cls.primary_color
 
         MDIconButton:
             id: button_1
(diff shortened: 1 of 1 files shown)


podraco (commented):
The property idea is a good one, but, the best practice is to avoid any kv rule, since they excecute even when not needed or are overwroten

podraco:
I can't say it's a definitive fix for the class, I'll need to test it first

PR #501, opened 2020-08-31
WIP - Buttons.py mayor bug fix and enhancements

podraco:
#### Pyproject
Pyproject.toml had misspelled spacing as a property of isort.

#### Buttons.py
the structure of every button has changed, now the base classes are now compatible with Base button instead to a determined kind of button.

Rectangular like button now depends on base button and either background or outline base class

The icon, Background and Outline of the buttons are now a behavior like class that allows the button class to show the specific canvas instruction

Background, outline and icon now uses a canvas instruction group, this allows to the developers to update the canvas instruction without the problem of mixing up the texture instructions, keeping them isolated from any other canvas event.

The Color Theme bug has been fixed, this was due to an over overwriting of the background color event that overrode every button when the main theme is changed. Even if the button's color is custom.

Option property had been added to control the behavior of a button, this allows to effectively control the color scheme of the button's: text, background, outline and even the icon color.

The new outline behavior class allows using rounded, sharp and bevel corners based in a unique property (radius)
the radius property is bounded to a maximum of half the height of the button, this is meant to comply with material IO description of that do and don't of a button shape

the button's corner style affects both background and outline shape since both rely on the same property

The button now has private properties that keeps track of the user input and behavior input. This allows to changes and restore the visuals of a button in a determined event. Like on disabled or on theme color of any property

The button's Label, Icon, Background and Outline are now theme-able, the base theme of a button is it's label's text
this means that icon and outline can be linked to the current text color by setting it's color theme to Text.

The only exception is the background since the text and the background of a button must never be the same color

All buttons have a container that's inside the anchor layout. this container keeps the icon, label and other element's needed to draw the button correctly, this container changes it's size depending on the elements and spacing

All buttons now have a padding property that only affects the button's container. if the container has an icon and a label, the spacing will separate them otherwise, the spacing has no effect at all

The documentation is being check to allow to show the base button docs

On disabled background bug is now fixed. The buttons that shouldn't have a background where showing a background when the property disabled is True. This was fixed due to the use of shaped_background_behaivor that rely on the property _is_filled
if this property is False the button's background canvas is clear and no longer processed.

#### New button behavior

This is an example of the current revision of the button's  test app. 
7V7mJKHeVp

#### Old button behavior:

active with custom background colors (they're not used and overwritten by the theme color)
SZc6Diikms

Disabled button
9vHL2v6VZo

PR #546, opened 2020-09-25
Fixed #545 by checking against self._current_button_color

### LINKED ISSUE #545: Cant set a button md_bg_color back to primary_color
### Description of the Bug

When you create a `MDRaisedButton` with a different color than `theme_cls.primary_color`, you can't set it back to the primary_color. The issue is caused by the following code in kivymd.uix.button:
```python
    def on_md_bg_color(self, instance, value):
        if value != [0.0, 0.0, 0.0, 0.0]:
            self._current_button_color = value
```
It prevents setting the color to the primary_color.

### Code and Logs

```python
from kivy.clock import Clock
from kivy.lang import Builder
from kivymd.app import MDApp
from kivymd.uix.button import MDRaisedButton
from kivymd.uix.screen import MDScreen

KV = '''
MyScreen
'''


class MyScreen(MDScreen):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.theme_cls = MDApp.get_running_app().theme_cls
        self.button = MDRaisedButton(text='CANCEL')
        self.button.md_bg_color = self.theme_cls.error_color
        self.add_widget(self.button)
        Clock.schedule_once(self.change_button, 10)

    def change_button(self, *_args, **_kwargs):
        self.button.text = "OK"
        self.button.md_bg_color = self.theme_cls.primary_color


class Test(MDApp):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def build(self):
        return self.screen

    def set_item(self, instance):
        # print(instance.text,"----------this is instance")
        gender = instance.text
        mm = self.root.ids.drop_item.set_item(gender)
        self.root.ids.drop_item.text = gender
        self.smenu.dismiss()


Test().run()
```

## Versions

* OS: any
* Python: 3.7
* Kivy: 1.11.1
* KivyMD: master

dolfandringa:
### Description of Changes
When setting a button's `md_bg_color` check against `self._current_button_color` instead of `self.theme_cls.primary_color`.

This fixes issue #545

### DIFF
--- a/kivymd/uix/button.py
+++ b/kivymd/uix/button.py
@@ -895,7 +895,7 @@ class BaseButton(
             self._current_text_color = value
 
     def on_md_bg_color(self, instance, value):
-        if value != self.theme_cls.primary_color:
+        if value != self._current_button_color:
             self._current_button_color = value
 
     def on_disabled(self, instance, value):


podraco:
Buttons api if now on a pull request, check the current changes

kivymd/uix/button.py
```diff
@@ -895,7 +895,7 @@ def on_text_color(self, instance, value):
             self._current_text_color = value
 
     def on_md_bg_color(self, instance, value):
-        if value != self.theme_cls.primary_color:
+        if value != self._current_button_color:
             self._current_button_color = value
```
podraco, line 900: This code does not solve any bug, since it will make the button lose its current color and be replaced by the theme assigned one.

podraco (commented):
I can't approve this pull request. sorry.

kivymd/uix/button.py
dolfandringa: Huh? I don't understand. Fixing replacing the background color is exactly the intention of the PR. Currently you can only replace the background color of a button in some cases, but not in others. 

My PR fixes the scenario where this works:
```
button = MDIconButton(text="OK")
button.md_bg_color = app.theme_cls.primary_color
# do something
button.text = "ERROR"
button.md_bg_color = app.theme_cls.error_color
```
But this doesn't:
```
button = MDIconButton(text="CANCEL")
button.md_bg_color = app.theme_cls.error_color
# do something
button.text = "OK"
button.md_bg_color = app.theme_cls.primary_color
```

podraco: Try changing the theme of the app, and you'll see the problem, as well others too.

Ideally, you should be able to assign a theme for the buttons, so the app will update all your UI to your theme color scheme.

For example, you should be able to set button.theme_button_color = "Error", so it will change it's color, internally.
And the md_bg_color won't interfere with the actual modification.

there's already a pr that solves this and other problems with the buttons

check #501 for more info

PR #570, opened 2020-10-06
Backdrop movement reference update

dwmoffatt:
### Description of Changes
- Added open_up_to check so the user can't cause not intended backdrop transitions
- Changes default open reference to use header_button and front_layer height

The above mention changes fix the problem of when you open the backdrop all the way you get part of whatever is put in the front layer. What is shown should just the the space where the header goes. Also this stops the user from being able to set the open_up_to direction to be to far below the screen or above 0 reference which shows the background, both are undesirable behaviors.

Implementation was tests with header property = True & False

### DIFF
--- a/kivymd/uix/backdrop.py
+++ b/kivymd/uix/backdrop.py
@@ -124,7 +124,6 @@ __all__ = (
 from kivy.animation import Animation
 from kivy.clock import Clock
 from kivy.lang import Builder
-from kivy.metrics import dp
 from kivy.properties import (
     BooleanProperty,
     ListProperty,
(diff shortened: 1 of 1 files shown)


dwmoffatt:
There is an issue with this PR. 

Working it out right now

dwmoffatt:
Issue has been resolved, was problem with test script. 

PR is good to be reviewed now. Tested with header = True and header = False

kivymd/uix/backdrop.py
```diff
@@ -328,7 +328,15 @@ def open(self, open_up_to=0):
             self.close()
             return
         if open_up_to:
-            y = open_up_to
+            # added open_up_to check
+            if open_up_to < (
+                self.ids.header_button.height - self.ids._front_layer.height
+            ):
+                y = self.ids.header_button.height - self.ids._front_layer.height
```
HeaTTheatR, line 335: This check will always be false.

dwmoffatt: No it will not. If you do .open(-1000) or some large value that is greater then the size of the _front_layer it goes in there and limits bad animations. In this instance it is the front_layer completely falling off the screen.

Issue #585, opened 2020-10-14
MDTextField multiline bug


allexeyj:
### Description of the Bug

I took that line(
"[INFO   ] [GL          ] NPOT texture support is available

"
) from log, and when i copy this line to the field, i get multiline, however i declared multiline = False parameter

### Code and Logs

```python
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
BoxLayout:
    padding: "50dp"

    MDTextField:
        multiline: False
        hint_text: "Helper text on error (press 'Enter')"
        pos_hint: {"center_y": .5}
'''


class Test(MDApp):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.screen = Builder.load_string(KV)

    def build(self):
        return self.screen


Test().run()
```

### Screenshots

![Test 14 10 2020 21_43_46](https://user-images.githubusercontent.com/71039746/96031789-7eb24f00-0e66-11eb-9ff5-3a338c0c139a.png)

### Versions

* OS: win10
* Python: 3.7
* Kivy: last version
* KivyMD: master branch

HeaTTheatR:
Fix - https://github.com/kivymd/KivyMD/pull/587/commits/07569e5a0fa7d5f9263cf58cbadfe89d09b225e9

PR #587, opened 2020-10-15
No newline if multiline False in MDTextField

shashi278:
Fixed bug pointed out in #585 
For a single-line textfield, if a pasted text contains a newline character, it should be replaced with a space.

### DIFF
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -405,8 +405,8 @@ With right icon
 
 __all__ = ("MDTextField", "MDTextFieldRect", "MDTextFieldRound")
 
-import sys
 import re
+import sys
 
 from kivy.animation import Animation
 from kivy.lang import Builder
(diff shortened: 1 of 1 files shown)


kivymd/uix/textfield.py
```diff
@@ -426,6 +426,8 @@ def build(self):
 from kivymd.theming import ThemableBehavior
 from kivymd.uix.label import MDIcon
 
+import re
+
```
HeaTTheatR: Thanks for these changes!
But could you move the modules of the standard library up:

Снимок экрана 2020-10-15 в 18 45 33

HeaTTheatR (changes requested):

kivymd/uix/textfield.py
shashi278: Sure.

HeaTTheatR (changes requested):
Also notice the following message:

Снимок экрана 2020-10-15 в 18 58 48

Follow the instructions on the link - https://github.com/kivymd/KivyMD#setup-environment

Issue #583, opened 2020-10-13
Dialog is not resizing with window


BeardyBarber:
### Description of the Bug

Dialog size in KivyMd is not changing when resizing window.
This works when opening and resizing window with dialog open and resizing before opening the dialog.

*Reproduction Steps:*
1. Run kitchen_sink demo APP
2. Click "Dialogs" in Menu
3. Click any button to open dialog
4. Resize down window

*Alternative Steps:*  
1. Run kitchen_sink demo APP
2. Resize down window
3. Click "Dialogs" in Menu
4. Click any button to open dialog

*Results:*
Dialog displayed is cut on the edges.

*Expected results:*
Dialog is resized to window width.
 
### Code and Log

USED KITCHEN_SINK DEMO APP FOR TESTING

### Screenshots
Run and open dialog:
![image](https://user-images.githubusercontent.com/20884974/95911432-c58a4100-0da1-11eb-8d12-fc6150c9f7c5.png)
After resizing:
![image](https://user-images.githubusercontent.com/20884974/95911465-d20e9980-0da1-11eb-8e4d-7352ca31097d.png)

### Versions
Tested on (but probably works on others too)
* OS: Windows
* Python: v3.8.5 (tags/v3.8.5:580fbb0)
* Kivy: v2.0.0rc3, git-20c14b2, 20200615
* KivyMD: 0.104.2.dev0, git-Unknown, 2020-10-12

HeaTTheatR:
Fix https://github.com/kivymd/KivyMD/pull/588

PR #588, opened 2020-10-15
Resizable width for MDDialog

shashi278:
As pointed out in #583, the width of `MDDialog` now resizes with the window.
For the resized width I've arbitrarily given left and right magin of `dp(24)` so that it doesn't look so close and compact with the device width. Also, I've restricted the dialog width to go beyond a certain limit.

Here's how it looks when resized:
Screenshot from 2020-10-15 23-42-23


kulothunganug:
i have an problem (in master branch)
the dialog width is only changes if the widow is resized

error

PR #606, opened 2020-10-24
Fix #605

### LINKED ISSUE #605: Weird behavior of MagicBehavior
### Description of the Bug

I noticed there is a problem with using relativelayout and magicbehavior . This bug also affects MDSwiper inside demo app. 

### Code and Logs

```python
from kivymd.app import MDApp

kv = """

<MyButton@MDRaisedButton+MagicBehavior>

<MyItem@RelativeLayout>
    MyButton:
        text: 'press'
        pos_hint: {'center_x': .5, 'center_y': .5}

Screen:

    BoxLayout:
        spacing: dp(5)

        MyItem:

        MyItem:

        MyItem:

        MyItem:

        """

class Main(MDApp):

    def build(self):
        return Builder.load_string(kv)

Main().run()

```

### Screenshots

ezgif com-video-to-gif(1)

quitegreensky:
### Description of Changes
This is a fix for #605 . Turned out clearing canvas.after inside ripplebehavior causes this issue. This fix will prevent from canvas.after being cleared every time. 

```python
    def anim_complete(self, *args):
        self._doing_ripple = False
        self._finishing_ripple = False
        self._fading_out = False
        self.canvas.after.clear()
```
### Screenshots

ezgif com-video-to-gif(2)

### DIFF
--- a/kivymd/uix/behaviors/ripplebehavior.py
+++ b/kivymd/uix/behaviors/ripplebehavior.py
@@ -201,6 +201,7 @@ class CommonRipple(object):
     _finishing_ripple = BooleanProperty(False)
     _fading_out = BooleanProperty(False)
     _no_ripple_effect = BooleanProperty(False)
+    _canvas_after_loaded = False
 
     def lay_canvas_instructions(self):
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


quitegreensky:
I noticed this fix breaks some other parts of the ripplebehavior. Not clearing canvas.after will make issues when trying to resize the screen. Reverting this commit is better idea until i find a proper fix :(

PR #610, opened 2020-10-25
Added radius option in picker

BananaMasterz:
### Description of Changes
Added the option to change the radius on the `MDDatePicker`,  `MDThemePicker `as well as `MDTimePicker `(works both in portrait and landscape mode).
You use it like this:

`MDDatePicker(self.set_date, radius=[dp(9), dp(9), dp(9), dp(9)]).open()`

Also added transparent background for the modalview

### Screenshots

image
image
image

### DIFF
--- a/kivymd/uix/picker.py
+++ b/kivymd/uix/picker.py
@@ -645,14 +645,16 @@ Builder.load_string(
     canvas:
         Color:
             rgba: self.theme_cls.bg_light
-        Rectangle:
+        RoundedRectangle:
             size: (dp(270), dp(335))
             pos: (root.pos[0], root.pos[1] + root.height - dp(335) - dp(95))
+            radius: [dp(0), dp(0), root.radius[2], root.radius[3]]
         Color:
             rgba: self.theme_cls.primary_color
-        Rectangle:
+        RoundedRectangle:
             size: (dp(270), dp(95))
             pos: (root.pos[0], root.pos[1] + root.height - dp(95))
+            radius: [root.radius[0], root.radius[1], dp(0), dp(0)]
         Color:
             rgba: self.theme_cls.bg_dark
         Ellipse:
@@ -691,6 +693,7 @@ class MDTimePicker(
     ThemableBehavior, FloatLayout, ModalView, RectangularElevationBehavior
 ):
     time = ObjectProperty()
+    radius = ListProperty([0, 0, 0, 0])
     """
     Users method. Must take two parameters:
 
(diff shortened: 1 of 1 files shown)


kivymd/uix/picker.py
```diff
@@ -218,9 +219,13 @@ def show_theme_picker(self):
                 (root.pos[0], root.pos[1] + root.height - dp(96))\
                 if self.theme_cls.device_orientation == 'portrait'\
                 else (root.pos[0], root.pos[1] + root.height - dp(304))
+            radius: [root.cal_radius,root.cal_radius,dp(0),dp(0)] \
+                    if self.theme_cls.device_orientation == 'portrait'\
+                    else [root.cal_radius,dp(0),dp(0),root.cal_radius]
```
HeaTTheatR: Use one space after the comma.

kivymd/uix/picker.py
```diff
@@ -229,6 +234,9 @@ def show_theme_picker(self):
                 (root.pos[0], root.pos[1] + root.height - dp(96) - (dp(484) - dp(96)))\
                 if self.theme_cls.device_orientation == 'portrait'\
                 else (root.pos[0] + dp(168), root.pos[1])
+            radius: [dp(0),dp(0),root.cal_radius,root.cal_radius] \
+                    if self.theme_cls.device_orientation == 'portrait'\
+                    else [dp(0),root.cal_radius,root.cal_radius,dp(0)]
```
HeaTTheatR: Use one space after the comma.

kivymd/uix/picker.py
```diff
@@ -448,6 +456,7 @@ class MDDatePicker(
     _sel_day_widget = ObjectProperty()
     cal_list = None
     cal_layout = ObjectProperty()
+    cal_radius = NumericProperty(0)
```
HeaTTheatR: You do not need to use the prefix `cal` for the `radius` parameter.

Why did you add the `radius` parameter only for the `MDDatePicker` class. I think it will be more logical if we make a common class `BaseDialog` (https://github.com/kivymd/KivyMD/blob/master/kivymd/uix/dialog.py#L180) for the `MDTimePicker`, `MDDatePicker`, `MDThemePicker` classes and specify the radius parameter in the `BaseDialoe` class.

HeaTTheatR (changes requested):

kivymd/uix/picker.py
BananaMasterz: `radius = ListProperty([7, 7, 7, 7])` if you mean this I can't use that because I only change 2 corners everytime example: `radius: [x, x, 0, 0]` or `radius: [x, 0 0, x]` etc because calendar is 2 different canvas's. The radius in BaseDialog returns all 4 corners radius. Correct me if I didn't understand correctly. I agree it would be better to have one radius for time, date and theme pickers. I didn't think of that.

HeaTTheatR: Yes, we will not be able to inherit from one `BaseDialog`. In this case, we need to add the `radius` parameter to all three classes `MDTimePicker`, `MDDatePicker`, `MDThemePicker`.

BananaMasterz: actually I found a way, check the last commit.

HeaTTheatR: I meant that we have to add the radius parameter to the rest of the classes - `MDTimePicker`, `MDThemePicker`.

BananaMasterz: Done. Let me know if that's better.

PR #625, opened 2020-10-31
fix typo

kulothunganug:


kivymd/uix/textfield.py
```diff
@@ -1036,7 +1036,7 @@ def on_line_color_focus(self, *args):
         if self.color_mode == "custom":
             self._update_colors(self.line_color_focus)
 
-    def on__hint_text(self, instance, value):
+    def on_hint_text(self, instance, value):
```
HeaTTheatR, line 1039: This is not a typo - https://github.com/kivy/kivy/blob/master/kivy/uix/textinput.py#L2633

HeaTTheatR (changes requested):

kivymd/uix/textfield.py
kulothunganug: Does it a mistake in typing or it's already correct?

HeaTTheatR: This is an overridden built-in method of the Kivy library textinput widget.

kulothunganug: :+1:

PR #692, opened 2020-11-26
🆙date chip-Demo

kulothunganug:


demos/kitchen_sink/libs/kv/chips.kv
```diff
@@ -35,33 +35,33 @@
 
                     MDChip:
                         label: "Coffee"
-                        color: .4470588235118, .1960787254902, 0, 1
+                        color: (.4470588235118, .1960787254902, 0, 1)
```
HeaTTheatR: You don't need to use parentheses, it makes the code difficult to read

HeaTTheatR (changes requested):

PR #698, opened 2020-11-30
 fix NameError: name 'Factory' is not defined on line 97.

zylo117:
Fix name error on line 97 because when you are using eval, some libraries are not imported.

I think pycharm removes unused import on optimize imports, and it won't care if you are using `eval()`.

### Description of Changes
* add an import


HeaTTheatR:
@zylo117
Thank you for helping to improve the KiwiMD library!
The import you specified is missing from the master branch.

Issue #697, opened 2020-11-29
MDTab bugs


kulothunganug:
### Description of the Bug

# bug when using `tab_indicator_type: "fill"`
![tab-fill](https://user-images.githubusercontent.com/63696279/100534814-f1ba2e00-3238-11eb-8ab1-4c3c6a8c80e5.gif)

```py
from kivy.lang import Builder
from kivy.uix.floatlayout import FloatLayout

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase

KV = '''
BoxLayout:
    orientation: "vertical"

    MDToolbar:
        title: "Example Tabs"

    MDTabs:
        id: tabs
        tab_indicator_anim: True
        tab_indicator_type: "fill"
        on_tab_switch: app.on_tab_switch(*args)


<Tab>:

    MDLabel:
        id: label
        text: "Tab 0"
        halign: "center"
'''


class Tab(FloatLayout, MDTabsBase):
    '''Class implementing content for a tab.'''


class Example(MDApp):
    def build(self):
        return Builder.load_string(KV)

    def on_start(self):
        for i in range(20):
            self.root.ids.tabs.add_widget(Tab(text=f"Tab {i}"))

    def on_tab_switch(
        self, instance_tabs, instance_tab, instance_tab_label, tab_text
    ):
        '''Called when switching tabs.

        :type instance_tabs: <kivymd.uix.tab.MDTabs object>;
        :param instance_tab: <__main__.Tab object>;
        :param instance_tab_label: <kivymd.uix.tab.MDTabsLabel object>;
        :param tab_text: text or name icon of tab;
        '''

        instance_tab.ids.label.text = tab_text


Example().run()
```

# without `tab_indicator_anim: True` this bug applies to all type of `tab_indicator_type`
![tab-bug-2](https://user-images.githubusercontent.com/63696279/100534736-27aae280-3238-11eb-9873-43385d3bedac.gif)

```py
from kivy.lang import Builder
from kivy.uix.floatlayout import FloatLayout

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase

KV = '''
BoxLayout:
    orientation: "vertical"

    MDToolbar:
        title: "Example Tabs"

    MDTabs:
        id: tabs
        # tab_indicator_anim: True
        tab_indicator_type: "fill"
        on_tab_switch: app.on_tab_switch(*args)


<Tab>:

    MDLabel:
        id: label
        text: "Tab 0"
[...]

PR #703, opened 2020-12-03
Fixed issues with indicator_type

shashi278:
Fixed issues with `tab_indicator_type` as mentioned in #697

### DIFF
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -466,7 +466,6 @@ Switching the tab by name
 __all__ = ("MDTabs", "MDTabsBase")
 
 from kivy.clock import Clock
-from kivy.graphics import RoundedRectangle
 from kivy.lang import Builder
 from kivy.properties import (
     AliasProperty,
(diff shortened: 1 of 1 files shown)


shashi278:
This is a better fix without involving clocks. I'd suggest you to merge this.

HeaTTheatR:
https://github.com/kivymd/KivyMD/commit/f5ceaa0f34097817cf65ed858a1a234ac7195331

shashi278:
> [f5ceaa0](https://github.com/kivymd/KivyMD/commit/f5ceaa0f34097817cf65ed858a1a234ac7195331)

I saw those changes that's why I said this one is better because it doesn't assume 2 sec delay in displaying indicator

HeaTTheatR:
@shashi278 Your changes will contain errors. When resizing the window ... Not required `_trigger_update_tab_indicator `method ... and so on

shashi278:
But this method was already there. I didn't introduce it. All I'm saying is, if we're talking about the indicator then we should not make it available using Clock. Plus, if there's any function we don't need anymore, we can simply remove it.

HeaTTheatR:
@shashi278 Try to pull requests with my changes and then add your own if you think my changes are not very good.

PR #705, opened 2020-12-03
Fixed issues with indicator_type

shashi278:
### Description of Changes
Removing `on_x` and `on_width` causes bug in the indicator so not removing them.

### DIFF
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -466,7 +466,6 @@ Switching the tab by name
 __all__ = ("MDTabs", "MDTabsBase")
 
 from kivy.clock import Clock
-from kivy.graphics import RoundedRectangle
 from kivy.lang import Builder
 from kivy.properties import (
     AliasProperty,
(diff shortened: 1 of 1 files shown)


kivymd/uix/tab.py
```diff
@@ -534,6 +533,8 @@ def switch_tab(self):
     color:
         self.text_color_active if self.state == 'down' \
         else self.text_color_normal
+    on_x: self._trigger_update_tab_indicator()
+    on_width: self._trigger_update_tab_indicator()
```
HeaTTheatR, line 537: Why are you calling these methods?

kivymd/uix/tab.py
```diff
@@ -635,6 +640,12 @@ def on_texture(self, widget, texture):
             self.width = texture.width
             self.min_space = self.width
 
+    def _trigger_update_tab_indicator(self):
+        # update the position and size of the indicator
+        # when the label changes size or position
+        if self.state == "down":
+            self.tab_bar.update_indicator(self.x, self.width)
+
```
HeaTTheatR, line 648: This method does nothing.

HeaTTheatR (changes requested):

shashi278:
@HeaTTheatR Any specific reason?

Issue #723, opened 2020-12-18
Strange Behavior when Adding/Removing Widgets


miniparth24:
### Description of the Bug

I'm trying to add lot of boxlayout widgets to a scrolls view and when I refresh the screen I'm trying to clear them all and add new sets of them, but it takes longer and longer every time I refresh to add same amount of widgets. Seems like clear_widget() clears the screen but it doesn't reclaim memory?

### Code and Logs

```python

from kivy.lang import Builder
from kivymd.app import MDApp
from kivy.properties import StringProperty
from kivymd.uix.boxlayout import MDBoxLayout
from datetime import datetime
from kivy.clock import Clock
import sys, os, psutil

kv = """
<hourly_banner>
    orientation: "vertical"

    MDSeparator:

    MDBoxLayout:
        size_hint: 0.95, 1
        pos_hint: {"center_x": .5}
        orientation: "horizontal"
        
        MDLabel:
            size_hint_x: 0.475
            text: root.weather_code
            font_size: self.height * 0.275
            halign: "center"

        MDBoxLayout:

            size_hint_x: 0.2
            MDIcon:
                icon: root.percip_icon
                font_size: self.height*0.375
            MDLabel:
                text: root.percip_percent
                font_size: self.height*0.275
                halign: "center"
    MDSeparator:
    
    
Screen:

    MDScrollViewRefreshLayout:
        id: refresh_layout_hourly
        refresh_callback: app.refresh_callback_hourly
        root_layout: root
        pos_hint: {"center_x": .5}


        MDGridLayout:
            id: hourly_grid
            cols: 1
            adaptive_height: True
            padding: "10dp", "0dp"
            row_default_height: "50dp"
            row_force_default: True
"""

class hourly_banner(MDBoxLayout):
    percip_percent = StringProperty()
    weather_code = StringProperty()
    percip_icon = StringProperty()


class Test(MDApp):

    def build(self):
        return Builder.load_string(kv)

    def on_start(self):
        self.setup_screen()

    def setup_screen(self):

        hourly_grid = self.root.ids['hourly_grid']

        print("Before clear Memory = {0}".format(psutil.Process(os.getpid()).memory_info().rss / 1024 ** 2))
        print("Ref before clear = {0}".format(sys.getrefcount(hourly_banner)))
        self.root.ids['hourly_grid'].clear_widgets()
        print("After clear Memory = {0} ".format(psutil.Process(os.getpid()).memory_info().rss / 1024 ** 2))
[...]

PR #739, opened 2020-12-28
Fixed property unbinding from theme_cls in MDLabel

quitegreensky:
#723 This is a suggestion about how we can fix unbinding from `theme_cls`. I have managed to unbind callbacks using `dec_disabled`, which gets triggered whenever a widget gets removed. If this solution is fine to you, i will fix this issue in other widgets as well.

### DIFF
--- a/kivymd/uix/label.py
+++ b/kivymd/uix/label.py
@@ -306,12 +306,11 @@ class MDLabel(ThemableBehavior, Label, MDAdaptiveWidget):
 
     parent_background = ListProperty(None, allownone=True)
 
-    _currently_bound_property = {}
-
     can_capitalize = BooleanProperty(True)
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@quitegreensky Does this solve the existing problem?

quitegreensky:
@HeaTTheatR Yes it does but this problem still exists everywhere that `self.theme_cls.bind()` is used . what do you think about adding `self._currently_bound_property `  and `dec_disabled` to `ThemableBehavior` as a default behavior? I tried and it works

HeaTTheatR:
> @HeaTTheatR Yes it does but this problem still exists everywhere that `self.theme_cls.bind()` is used . what do you think about adding `self._currently_bound_property ` and `dec_disabled` to `ThemableBehavior` as a default behavior? I tried and it works

Yes, I thought about that too. It seems that we should add this to the `ThemableBehavior` class.

PR #749, opened 2021-01-04
Removed unwanted font files.

kulothunganug:
### Description of Changes
* Removed unwanted font files because `Kivy` already have these fonts and registered those.


kivymd/font_definitions.py
```diff
@@ -12,13 +12,6 @@
 from kivymd import fonts_path
 
 fonts = [
-    {
-        "name": "Roboto",
-        "fn_regular": fonts_path + "Roboto-Regular.ttf",
-        "fn_bold": fonts_path + "Roboto-Bold.ttf",
-        "fn_italic": fonts_path + "Roboto-Italic.ttf",
-        "fn_bolditalic": fonts_path + "Roboto-BoldItalic.ttf",
-    },
```
HeaTTheatR, line 21: Actually, these fonts are used.

HeaTTheatR (changes requested):

kivymd/font_definitions.py
kulothunganug: these fonts are already exists in https://github.com/kivy/kivy/tree/master/kivy/data/fonts so we don't want to use separately.

HeaTTheatR: Have you compared these fonts? They differ.

kulothunganug: No, How do i compare them?

kulothunganug: Mostly it would be same.

kulothunganug: refs
https://github.com/kivy/kivy/blob/5e1311913722387242e1a3a48267333843a0e4a9/kivy/config.py#L867
https://github.com/kivy/kivy/blob/5e1311913722387242e1a3a48267333843a0e4a9/kivy/core/text/__init__.py#L94

HeaTTheatR:
Kivy and KivyMD fonts are very different. We have no right to delete them.

PR #763, opened 2021-01-09
Fix Docs

kulothunganug:
### Description of Changes
* Fix typo (a and an)
* Removed `get_color_from_hex`
> When handling with `ColorProperty` we can directly give a hex color or rgba it would recognize automatically.
> ex: "FFFFFF", "#FFFFFF", (1, 1, 1, 1)

#### See files by commit to get a clear understand

### DIFF
--- a/kivymd/theming.py
+++ b/kivymd/theming.py
@@ -708,7 +708,7 @@ class ThemeManager(EventDispatcher):
     def _set_ripple_color(self, value):
         self._ripple_color = value
 
-    _ripple_color = ColorProperty(get_color_from_hex(colors["Gray"]["400"]))
+    _ripple_color = ColorProperty(colors["Gray"]["400"])
     """Private value."""
 
     ripple_color = AliasProperty(
--- a/kivymd/theming.py
+++ b/kivymd/theming.py
@@ -731,7 +731,7 @@ class ThemeManager(EventDispatcher):
     """
     Device orientation.
 
-    :attr:`device_orientation` is an :class:`~kivy.properties.StringProperty`.
+    :attr:`device_orientation` is a :class:`~kivy.properties.StringProperty`.
     """
 
     def _get_standard_increment(self):
@@ -848,7 +848,7 @@ class ThemeManager(EventDispatcher):
 
     .. image:: https://github.com/HeaTTheatR/KivyMD-data/raw/master/gallery/kivymddoc/font-styles.png
 
-    :attr:`font_styles` is an :class:`~kivy.properties.DictProperty`.
+    :attr:`font_styles` is a :class:`~kivy.properties.DictProperty`.
     """
 
     def __init__(self, **kwargs):
@@ -874,7 +874,7 @@ class ThemableBehavior(EventDispatcher):
     """
     ``True`` if device is ``iOS``.
 
-    :attr:`device_ios` is an :class:`~kivy.properties.BooleanProperty`.
+    :attr:`device_ios` is a :class:`~kivy.properties.BooleanProperty`.
     """
 
     opposite_colors = BooleanProperty(False)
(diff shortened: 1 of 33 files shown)


kivymd/theming.py
```diff
@@ -40,7 +40,6 @@
 
 from kivymd import images_path
 from kivymd.color_definitions import colors, hue, palette
-from kivymd.font_definitions import theme_font_styles  # NOQA: F401
```
HeaTTheatR, line 43: It cannot be deleted (Flake8).

HeaTTheatR (changes requested):

kivymd/uix/backdrop.py
```diff
@@ -218,7 +218,7 @@ class MDBackdrop(ThemableBehavior, FloatLayout):
     """
     Padding for contents of the front layer.
 
-    :attr:`padding` is an :class:`~kivy.properties.ListProperty`
+    :attr:`padding` is a :class:`~kivy.properties.ListProperty`
```
HeaTTheatR, line 221: https://github.com/kivy/kivy/blob/master/kivy/uix/codeinput.py#L70

HeaTTheatR (changes requested):

kivymd/uix/dialog.py
```diff
@@ -181,7 +181,11 @@ def show_alert_dialog(self):
 
 
 class BaseDialog(ThemableBehavior, ModalView):
-    radius = ListProperty([7, 7, 7, 7])
+    radius = ListProperty(
+        [
+            dp(7),
+        ]
+    )
```
HeaTTheatR, line 188: This is an error - the radius is intended to change the value of all angles, not just one. The four-element list means that you can set different values for any angles.

HeaTTheatR (changes requested):

kivymd/theming.py
```diff
@@ -731,7 +731,7 @@ def _determine_device_orientation(self, _, window_size):
     """
     Device orientation.
 
-    :attr:`device_orientation` is an :class:`~kivy.properties.StringProperty`.
+    :attr:`device_orientation` is a :class:`~kivy.properties.StringProperty`.
```
HeaTTheatR, line 734: https://github.com/kivy/kivy/blob/master/kivy/uix/codeinput.py#L70

HeaTTheatR (changes requested):

kivymd/uix/backdrop.py
kulothunganug: Sorry I can't understand

kivymd/theming.py
kulothunganug: Does `theme_text_color` used in any other files?
like `from kivymd.theming import theme_font_styles`

kivymd/theming.py
kulothunganug: Sorry I can't understand

kivymd/theming.py
HeaTTheatR: If after the import the `# NOQA: F401` tag is set, then it is used and this import cannot be deleted.

kivymd/uix/dialog.py
kulothunganug: https://github.com/kivymd/KivyMD/blob/93d6823b3eb6cbd6574f81852536cc548846acad/kivymd/uix/menu.py#L905
https://github.com/kivymd/KivyMD/blob/93d6823b3eb6cbd6574f81852536cc548846acad/kivymd/uix/tooltip.py#L172

before this [commit](https://github.com/kivymd/KivyMD/commit/69680954753c9f2e7cc57504bb918940ea9af5a8#diff-d808ef41d27ee485ae6d3f04e2cb96c5d3774c2696ffbc3299d33b15e7cd3ab3) in mobile the edges are sharp
after the edges are looks same in all platforms

PR #768, opened 2021-01-12
Fix: Uncaught error in MDTabsBar

### LINKED ISSUE #767: Uncaught Error Condition in MDTabsBar
### Description of the Bug

There is a logic hole in the `tab_bar_autoscroll` method and I have a Tab that raises an uncaught error because of it. I do not know (yet) why this tab is triggering the problem, but regardless the error should not be uncaught. 

The but is in lines 861 - 868 of tabs.py:
```python
       if not dst:
            return

        if scroll_is_late and target.center_x > bound_left:
            x = lsx + dst

        elif not scroll_is_late and target.center_x < bound_right:
            x = lsx - dst
```

The conditions do not cover all possible result sets - and so x can be undefined.

I propose, and will PR, adding a default trapdoor:

```python
        if not dst:
            return

        if scroll_is_late and target.center_x > bound_left:
            x = lsx + dst

        elif not scroll_is_late and target.center_x < bound_right:
            x = lsx - dst
            
        else:
            return
```

In my erroneous Tab - this change means the only effect is that the tab does not autoscroll. Improved from the current complete crash.

### Versions

* OS: mac
* Python: 3.8.6
* Kivy: 2.0.0
* KivyMD: master

runette:
close #767 

### Description of Changes

Add a default trapdoor to the autoscroll function

### DIFF
--- a/kivymd/toast/androidtoast/androidtoast.py
+++ b/kivymd/toast/androidtoast/androidtoast.py
@@ -37,7 +37,6 @@ AndroidToast
 
     Test().run()
 """
-
 all = ("toast",)
 
 from android.runnable import run_on_ui_thread
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -873,6 +873,9 @@ class MDTabsBar(ThemableBehavior, RectangularElevationBehavior, MDBoxLayout):
         elif not scroll_is_late and target.center_x < bound_right:
             x = lsx - dst
 
+        else:
+            return
+
         x = boundary(x, 0.0, 1.0)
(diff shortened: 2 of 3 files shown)


runette:
This is the second PR I have submitted and the second time there have been lint errors that have nothing to do with code I have submitted. There is something badly wrong if there are errors in the Master code base.

These errors :
```
kivymd/toast/androidtoast/androidtoast.py:42:1: E402 module level import not at top of file
kivymd/toast/androidtoast/androidtoast.py:43:1: E402 module level import not at top of file
```

are nothing to do with anything I have submitted and I do not know enough about Android to solve them.

PR #755, opened 2021-01-06
#754

runette:
### Description of Changes
Changes to allow the palette colors to be overridden as discussed in #754 

Change `colors` in `ThemeManager` to `self.colors` and default to the standard palette.

To create custom colors, create your version of `color_definitions.py` and change one of the existing color definitions (it has to be one of the existing ones because the palette color names are hard coded.

PR #773, opened 2021-01-13
Fix for #772 - set_colors

### LINKED ISSUE #772: Define a courtesy method to set all theme colors
### Description of the Feature

The current process for setting up a palette (unless you are accepting the most basic defaults) gets very long winded and boilerplate-ey with 6 different method calls.

It would be good to create a courtesy method to set them all in one call.

I will submit a PR

runette:
closes #772 

### Description of Changes

Adds a method to `ThemeManager` that allows all of the theme color attributes to be set in one call.

Example and documentation included

### DIFF
--- a/kivymd/theming.py
+++ b/kivymd/theming.py
@@ -852,7 +852,17 @@ class ThemeManager(EventDispatcher):
     :attr:`font_styles` is an :class:`~kivy.properties.DictProperty`.
     """
 
-    def set_colors(self, primary_palette, primary_hue, primary_light_hue, primary_dark_hue, accent_palette, accent_hue, accent_light_hue, accent_dark_hue):
+    def set_colors(
+        self,
+        primary_palette,
+        primary_hue,
+        primary_light_hue,
+        primary_dark_hue,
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@runette You have not added any documentation that we talked about earlier.

runette:
I think you have different PRs mixed up - we have not discussed anything about this PR before and the method IS documented.

runette:
This is not in any way linked to #755. This is just a courtesy method to avoid you having to make 8 different calls if you need to set all of the theme attributes (i.e. the primary and accent colors with light and dark variants for each one).

kulothunganug:
I think using `**kwargs` will be good.

runette:
Too complicated. If the users want to set the attributes individually then
they can use the existing methods. Adding kwargs here would add as much
boilerplate as using the base methods. This method is just for users who
want to set the entire theme in one set and using positional is much
cleaner: `set_colors("Red', "50","500", "900", "Blue", "50", "500", "900")`
etc

PR #798, opened 2021-01-28
Fixed MDBottomAppBar notch transparency

quitegreensky:
### Description of Changes
Fixed the transparency issue in `MDBottomAppBar` by using `Mesh` and giving a custom curve to it.

### Screenshots

Untitled

### DIFF
--- a/kivymd/uix/toolbar.py
+++ b/kivymd/uix/toolbar.py
@@ -278,6 +278,8 @@ You can add MDTooltips to the Toolbar icons by ading a text string to the toolba
     `Components-Bottom-App-Bar <https://github.com/kivymd/KivyMD/wiki/Components-Bottom-App-Bar>`_
 """
 
+from math import cos, radians, sin
+
 from kivy.animation import Animation
 from kivy.clock import Clock
 from kivy.core.window import Window
@@ -320,8 +322,7 @@ Builder.load_string(
     canvas.after:
         PopMatrix
 
-
-<MDToolbar>
+<NotchedBox>
     size_hint_y: None
     height: root.theme_cls.standard_increment
     padding: [root.theme_cls.horizontal_margins - dp(12), 0]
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@quitegreensky 
Great job!
There are two questions: have you tested these changes on mobile?
Can the rounding be smoother? The screenshot shows that there are sharp corners where a smoother should be.

Снимок экрана 2021-01-29 в 00 39 31

quitegreensky:
@HeaTTheatR It looks smoother now. also i will try to build kitchen_sink for android and let you know if there is any problem.
Untitled

EDIT:
@HeaTTheatR 
On Android:
photo_2021-01-28_02-37-45

quitegreensky:
@HeaTTheatR  According to [spec](https://material.io/components/app-bars-bottom#anatomy) placing FAB outside of the bottom app bar is not recommended. so is it necessary to keep `Free-End` and `Free-Center` options for FAB?
Untitled

HeaTTheatR:
> @HeaTTheatR According to [spec](https://material.io/components/app-bars-bottom#anatomy) placing FAB outside of the bottom app bar is not recommended. so is it necessary to keep `Free-End` and `Free-Center` options for FAB?
> Untitled

Yes, there should be such features. Material Design does not recommend this, but it is possible to do so.

HeaTTheatR:
@quitegreensky 
Do you have any news on this pull request? Have you taken the time to test these changes on Android device?

quitegreensky:
@HeaTTheatR  This is my test results. I didn't notice any issues on my own device by now.

https://user-images.githubusercontent.com/55235027/106598398-8593c900-650c-11eb-8396-92f999a9d414.mp4

https://user-images.githubusercontent.com/55235027/106599079-6cd7e300-650d-11eb-994c-7c3847256400.mp4

HeaTTheatR:
@quitegreensky 
Excellent!
It seems to me that the distance between the FAB and the toolbar is very different ...

106016732-d0f34500-60d0-11eb-9c03-44cbef135d7a

HeaTTheatR:
@quitegreensky One more question. `Mesh` instructions require constant rendering of the screen. Do these changes provide a method to stop the updates of these instructions?

quitegreensky:
> 
> 
> @quitegreensky
> Excellent!
> It seems to me that the distance between the FAB and the toolbar is very different ...
> 
> 106016732-d0f34500-60d0-11eb-9c03-44cbef135d7a

@HeaTTheatR [spec](https://material.io/components/app-bars-bottom#behavior) hasn't defined the spacing between toolbar and FAB as i searched. I changed it to `dp(4)` for now which is correct i think. I also solved some other mathematical issues. but i don't understand your question about stopping the updates of `Mesh`. I've bound the method `_update_canvas` to certain events which will update the mesh only when needed. so i don't think if there is a constant rendering of screen.  
photo_2021-02-01_19-38-40

darpan5552:
image

I think notch should be a combination of 2 Bezier curves. Anyway, congrats for the work you put in.

HeaTTheatR:
@darpan5552 I think we'll come back to this question a little later...

quitegreensky:
@darpan5552 yes but its very slow yo calculate the points on a bezier without using something like numpy.

HeaTTheatR:
@quitegreensky So it will work faster with `NumPy`?

quitegreensky:
sina, [03.02.21 13:27]
> @quitegreensky So it will work faster with NumPy?

It's not a big problem for calculating circles to use pure python but in case of using bezier yes it's faster to use `numpy` but it will add another dependency only for this situation. of cource we can consider using `cython` which requires it's recipe to work.

HeaTTheatR:
@quitegreensky I think we need to create a new issue labeled `Enhancement` to come back to this issue in the future. Dependency `NumPy` for just one task is not a good solution.

quitegreensky:
@HeaTTheatR By using kivy's `Bezier` instruction to draw a bezier curve line, it already calculates all vertices we need but seems its not possible to somehow have access and return those points.

darpan5552:
@quitegreensky 
This widget has primary usage in mobile devices. So, user won't be changing size of screen now & then. In this context, bezier curves wouldn't be an extra load, rather it provides an elegant way for UI designers to make user do almost everything on a single screen.

Yes, i visited bezier.py documentation before making a comment here. Might be a modification required to make it return the points.

HeaTTheatR:
@quitegreensky, @darpan5552
Unfortunately, I cannot help with this problem as I have not yet delved into the `Mesh` instructions...

Issue #840, opened 2021-02-11
MDToolbar Floating action button error


podraco:
### Description of the Bug

When you set any toolbar, the shadow of the MDFloatingButtonAction button was modified by the toolbar class directly, thus forcing the shadow. even when the widget was not present.

thanks to #838 we can see that the shado of the button was always pressent. and animated. but behind the background instruction.

### Code and Logs
used kitchen sink demo from master.

### Screenshots
![Captura de pantalla de 2021-02-11 17-38-15](https://user-images.githubusercontent.com/5509325/107713451-e2238080-6c90-11eb-837a-5aefc0600f49.png)

This is just the report of the bug, this error is being fix.

PR #844, opened 2021-02-12
MDToolbar elevation fix

podraco:
### Description of Changes
Thanks for fixing this issue, i was already working on it. this part 
tho, takes back control of the elevation to the elevation behavior and 
avoids multiple actions when self.type is "top".

*. Run Pre-Commit 
*. removed unnecesary imports.


podraco:
This is a continuation for #840

Issue #851, opened 2021-02-14
MDTextField not validating on text change


BeardyBarber:
### Description of the Bug

MDTextField is not validating on input only when it lose or gain focus.

Reproduction steps for below code:
1. Run example (text field gain focus on start) 
2. Write something (helper text still visible)
3. Lose focus clicking on background (helper text disappear)

In my understanding when typing validation should be made after any change to text but now it is only made on focus in and out.

### Code and Logs

```from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDTextField:
    hint_text: "required = True"
    required: True
    helper_text_mode: "on_error"
    helper_text: "Enter text"
    focus: True
    pos_hint: {'center_x': .5, 'center_y':.5}
'''


class TestCard(MDApp):
    def build(self):
        return Builder.load_string(KV)


TestCard().run()

```

### Screenshots
![Peek 2021-02-14 13-03](https://user-images.githubusercontent.com/20884974/107883047-c7e5de80-6eec-11eb-9709-3fe8b766c61f.gif)

### Versions

* OS: MINT 20.1
* Python: 3.8.5
* Kivy: master (2021-02-14)
* KivyMD: master (2021-02-13)

HeaTTheatR:
@BeardyBarber Please describe your question in one message, attach a minimal code that may be run without additional manipulations, so developers can look at the working script and fix it. Also write what you expect from the script and what happened wrong. If needed, attach links or screenshots to explain us what you want to do.

BeardyBarber:
@HeaTTheatR I'm sorry, I edited my issue. Probably accidentally refreshed site after I filled the whole issue and didn't noticed it get cleared...

[Edit] Also the `error` attribute for MDTextField always return False even if it is empty, let me know if it needs another issue.

podraco:
MDTextField is based in kivy.uix.textinput, there it allows you to see what events are dispatched every time you type something inside.

BeardyBarber:
@podraco Can you tell me how can i get to these events and use this knowledge to my needs?

runette:
As far as I understand the issue - the fix for #858 contained in PR #859 also fixes this issue - see the following 

![Feb-15-2021 22-23-02](https://user-images.githubusercontent.com/2239795/107998228-8cc2d880-6fdc-11eb-97db-8ec32c640c35.gif)

BeardyBarber:
Working as expected, closing.

PR #831, opened 2021-02-10
Request a message update for the helper text

podraco:
Fix #824

resume: the __init__ call didnt call the update message event because 
the inner memory was set inside the super()__init__ call.

### Description of Changes
The class request a message update inside `__init__`

### Screenshots
```py
from kivymd.app import MDApp
from kivymd.uix.screen import MDScreen
from kivymd.uix.textfield import MDTextField


class Example(MDApp):
    def build(self):
        screen = MDScreen()
        screen.add_widget(
            MDTextField(
                hint_text="hint_text",
                helper_text="helper_text",
                helper_text_mode="on_focus",
                pos_hint={"center_x": .5, "center_y": .5},
                size_hint_x=0.5,
            )
        )
        return screen


Example().run()

```
#### Before
Captura de pantalla de 2021-02-09 18-13-42

#### Before
Captura de pantalla de 2021-02-09 18-13-32

PR #820, opened 2021-02-08
Update textfield.py

Neizvestnyj:
Now, when you set the `text`, `hint_text` goes up (text=self.set_text)

PR #859, opened 2021-02-15
Fix the problem with Textfield not responding to text set in Python

### LINKED ISSUE #858: Latest Update has Broken MDTextfield
### Description of the Bug

The latest updates to MDTextfield have stopped it working correctly if text is set in Python. The Hint text no longer animates. See example and gif below. This problem did not exist until about 11 days ago.

This is because the relevant function - that was `def on_text` has been changed to `def set_text` and so no longer automatically triggers when text is set after the Widget is created.

Someone raised a PR - #820 that was a fix - but you closed the PR. However, I am not sure that is the best fix. Far better just to use the functionality of Kivy and rename the function back to `on_text`. It can still be called from `def check_text` as it is currently.

### Code and Logs

```python
from kivy.lang import Builder
from kivy.properties import StringProperty
from kivymd.app import MDApp

KV = '''
BoxLayout:
    orientation: "vertical"
    BoxLayout:
        padding: "10dp"
        orientation: "horizontal"
    
        MDTextField:
            id: text_field_1
            hint_text: "Helper text"
            pos_hint: {"center_y": .5}
            text: app.text1
            
        MDTextField:
            id: text_field_2
            hint_text: "Helper text"
            pos_hint: {"center_y": .5}
            text: app.text2

    BoxLayout:
        orientation: "horizontal"
        Widget:
        MDRaisedButton:
            text: "Populate"
            on_release: app.populate()
        Widget:
    BoxLayout:
'''


class Test(MDApp):
    text1 = StringProperty()
    text2 = StringProperty()

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def build(self):
        return self.screen

    def populate(self):
        self.text1 = "One"
        self.text2 = "Two"


Test().run()
```

### Screenshots

Feb-15-2021 21-49-48

### Versions

* OS: 
* Python: 
* Kivy: 
* KivyMD:

runette:
fixes #858 

### Description of Changes
- Change the function `def set_text` back to `def on_text` so that it is triggered when the text is updated in python

### Screenshots

Using the same code as shown in issues #858 

Feb-15-2021 22-14-59

### DIFF
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -951,7 +951,7 @@ class MDTextField(ThemableBehavior, TextInput):
         self._set_msg(self, self.helper_text)
 
     def check_text(self, interval):
-        self.set_text(self, self.text)
+        self.on_text(self, self.text)
 
     def set_objects_labels(self):
         """Creates labels objects for the parameters
@@ -1116,7 +1116,7 @@ class MDTextField(ThemableBehavior, TextInput):
             elif self.color_mode == "custom":
                 self._update_colors(self.line_color_focus)
 
-    def set_text(self, instance, text):
+    def on_text(self, instance, text):
         self.text = re.sub("\n", " ", text) if not self.multiline else text
         if len(text) > 0:
             self.has_had_text = True
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -939,6 +939,7 @@ class MDTextField(ThemableBehavior, TextInput):
             _hint_lbl_font_size=self._hint_lbl.setter("font_size"),
             helper_text_mode=self._set_message_mode,
             max_text_length=self._set_max_text_length,
+            text=self.set_text,
         )
         self.theme_cls.bind(
             primary_color=self._update_primary_color,
@@ -951,7 +952,7 @@ class MDTextField(ThemableBehavior, TextInput):
         self._set_msg(self, self.helper_text)
 
     def check_text(self, interval):
-        self.on_text(self, self.text)
+        self.set_text(self, self.text)
 
     def set_objects_labels(self):
         """Creates labels objects for the parameters
@@ -1116,7 +1117,7 @@ class MDTextField(ThemableBehavior, TextInput):
             elif self.color_mode == "custom":
                 self._update_colors(self.line_color_focus)
 
-    def on_text(self, instance, text):
+    def set_text(self, instance, text):
         self.text = re.sub("\n", " ", text) if not self.multiline else text
         if len(text) > 0:
             self.has_had_text = True
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -1124,22 +1124,22 @@ class MDTextField(ThemableBehavior, TextInput):
             self._right_msg_lbl.text = f"{len(text)}/{self.max_text_length}"
         self._set_text_len_error()
         if self.error or self._text_len_error:
-            if self.focus:
-                self._anim_current_line_color(self.error_color)
-                if self.helper_text_mode == "on_error" and (
-                    self.error or self._text_len_error
-                ):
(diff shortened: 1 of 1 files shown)


runette:
Also fixes #851

runette:
Updated - added an additional commit to improve the handling of text added from Python

- If the text causes an error, then the error handling will now happen correctly
- If the text is valid - then the textfield will now be correctly formatted and colored as if the text was entered manually

HeaTTheatR (changes requested):
I deliberately renamed the `on_text` method to `set_text` because it fixes the error when setting the default text.

podraco:
If you use the common name, it can be rewritten by the front end developer
The (on_text) name it's probably going to be rewritten by the end user of kivymd.

runette:
> If you use the common name, it can be rewritten by the front end developer
> The (on_text) name it's probably going to be rewritten by the end user of kivymd.

Probably is 'probably' an extremely biased view! I think that you will find the vast majority of use cases will not override the `on_text` function.

Note that this does not apply to using `on_text:` in KV language since this creates a binding and does not override the function.

Any function CAN be overridden. `on_text` is 'probably' more likely to be overridden because that is the behaviour that people expect. That is a good thing - making code act as expected and easy to understand are both good things. Anyone who knows enough to sub class the MDTextField and over ride `on_text` should also know enough to `super` the base class function especially if they find they are getting unexpected behaviour.

Lets also be aware that until 10 days ago it WAS called `on_text` and I cannot find any example of anyone complaining. You could also apply the same argument to `on_focus` and `on_disabled` etc but they are still there.

However, ultimately it doesn't matter as much as getting this fixed - the fact that all of the TextFields in my app are now broken is a bit of a problem for me. So - I can change it to another name. It does mean accepting the change from the original PR #820.

runette:
> I deliberately renamed the `on_text` method to `set_text` because it fixes the error when setting the default text.

I am not sure which issue that is - since I cannot find any issue referring to "default text" or indeed any function or attribute on either kivy TextField or kivymd MDTextField called "default text". If the issue is the fix in #831 then that fix still works (as shown in the gif below running the test code for #831 against this PR) - since the fix is not the name of the function but the fact that it is called in `init`. However, changing the name from on_text does mean that it is no longer called whenever the text is changed - which it was before and the correct function of the TextField depends on that. If you want the function name changed, then the new function MUST bind to the on_text event - as per #820.

I don't think that changing the name of the function actually fixes anything per se. If you are saying that there is a problem with calling this function whenever the text changes then the issue is much deeper and more critical - since you have broken a whole part of the core functionality of the component.

Feb-16-2021 10-08-07

kivymd/uix/textfield.py
```diff
@@ -1116,7 +1117,7 @@ def on_disabled(self, *args):
             elif self.color_mode == "custom":
                 self._update_colors(self.line_color_focus)
 
-    def on_text(self, instance, text):
```
HeaTTheatR: I don't understand where you got this code from ... 
I fixed that a long time ago. And these changes are in the master branch.

Снимок экрана 2021-02-16 в 18 35 15

HeaTTheatR (changes requested):

kivymd/uix/textfield.py
runette: To be clear about one thing - this PR was created against the current Master branch and when you look on GH at the "files changed" view or the "review view" you are seeing a diff against the Master Branch so you can see what the changes ARE against the Master branch.

This PR in the current form (after the conversation above about function names) is made up of two main changes:

- line 942 binds the `set_text` method to the `on_text` event.
- in lines 1127 - 1143 remove the `if self.focus` test since changes that come from binding the text attribute to a StringProperty are most likely to happen when the Widget is NOT focussed and this test means that this can leave the Widget in an incorrect state.

Which of those two changes were you referring to?

HeaTTheatR:
@runette Your changes fix these problems, I checked. Thank.

PR #875, opened 2021-02-18
Fix for #874

### LINKED ISSUE #874: Problem with MDTabs and text_size causing incorrect sizing in Tab Labels
### Description of the Bug

This is a complicated one - so I will start with TL;DR:

> I have a problem with line 520 in `tabs.py` - it does not add anything that I can see and in a very specific circumstance it causes the Tab Label to be incorrectly sized.

Line 520 says this:
```
        if self.tab_bar.parent.allow_stretch else (self.width, None)
```

### Background

[The Fix for #534](https://github.com/kivymd/KivyMD/commit/3749028c9953238003f6d8473fd7e24d8670b52e) changed the Tab Label to be a subclass of MDLabel rather than Label (which is a good thing) and also added this code :

```
 on_tab_bar:
        self.text_size = (None, None) \
        if self.tab_bar.parent.allow_stretch else (self.width, None) 
```

I think this is wrong. The only reason for having `text_size` set is to control wrapping etc. For MDLabel, it makes sense. However, for MDTabs - as far as I can see it - there is no mode where you want the label text to wrap. So it makes no sense to set the `text_siz`e to anything. However, since MDLabel sets the `text_size` to `self.width` then it is necessary to over write`text_size` to `[None, None]` - but this should be done for all modes.

the logic that follows is to set the label width and label min_width from the texture size - which makes sense if `text-size` is not set. However, if `text_size` is set - then the texture size is set by the text_size - which we just set from `self.width` - so the whole process is circular!

### The problem

In normal operation - what I have described above does NOT cause any problems. However, it also does nothing useful which I have tested by running Kitchen sink and other apps without that line.

My main app - however- uses a non-standard font loaded in `theming`. For whatever reason - doing this means that the Label `self.width` is not set correctly until after the texture is created (there are notes to this effect in the kivy documentation). This means that line 520 sets an incorrect size - which then gets copied into the texture size.

This effect can be reproduced - but I have not put the code here since it is complicated and requires you to download some font files. Instead - I have shown the effect in the kitchen sink app. 

First with line 520

Feb-18-2021 01-24-02

Then without line 520

Feb-18-2021 01-24-40

### Versions

* OS: macos
* Python: 3.9
* Kivy: 2.0.0
* KivyMD: latest master

runette:
fixes #874 

### Description of Changes
* remove line 520 in `tabs.py` as discussed in #874

### DIFF
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -516,8 +516,7 @@ Builder.load_string(
         self.tab_bar.parent.carousel.current_slide.tab_label); \
         self._set_start_tab = True
     on_tab_bar:
-        self.text_size = (None, None) \
-        if self.tab_bar.parent.allow_stretch else (self.width, None)
+        self.text_size = (None, None)
     on_ref_press:
         self.tab_bar.parent.dispatch(\
         "on_ref_press",


podraco:
Don't remove anything until it's well tested. Please. 
The class by itself is quite complex.

HeaTTheatR:
> Don't remove anything until it's well tested. Please.
> The class by itself is quite complex.

I can only check this when I get back from work.

podraco:
the docs of MDTabs states the following: 
``` py
allow_stretch = BooleanProperty(True)
    """
    If False - tabs will not stretch to full screen.

    :attr:`allow_stretch` is an :class:`~kivy.properties.BooleanProperty`
    and defaults to `True`.
    """
```

podraco:
As a side note, if this change is approved, it's incomplete. Since it's only deleting where the memory is used, but not where the memory is  assigned.

runette:
> the docs of MDTabs states the following:
> 
> ```python
> allow_stretch = BooleanProperty(True)
>     """
>     If False - tabs will not stretch to full screen.
> 
>     :attr:`allow_stretch` is an :class:`~kivy.properties.BooleanProperty`
>     and defaults to `True`.
>     """
> ```

Whilst true, I am not sure about the point that you are making?

runette:
> As a side note, if this change is approved, it's incomplete. Since it's only deleting where the memory is used, but not where the memory is assigned.

This does not make sense. The change is about what value to set a kivy property. As such, it is not about memory allocation in anyway.

Before this change, the KV language sets the `text_size` to one of two values based on the value of `allow_stetch`. After this change the KV language sets `text_size` to the same value regardless of the value of `allow_stretch`.

podraco:
> > As a side note, if this change is approved, it's incomplete. Since it's only deleting where the memory is used, but not where the memory is assigned.
> 
> This does not make sense. The change is about what value to set a kivy property. As such, it is not about memory allocation in anyway.
> 
> Before this change, the KV language sets the `text_size` to one of two values based on the value of `allow_stetch`. After this change the KV language sets `text_size` to the same value regardless of the value of `allow_stretch`.

I'm referring to the text above. That memory is only used for that code line, if you remove that code line, you  should remove the other line as well, otherwise it will be left forgotten.

runette:
> > > As a side note, if this change is approved, it's incomplete. Since it's only deleting where the memory is used, but not where the memory is assigned.
> > 
> > 
> > This does not make sense. The change is about what value to set a kivy property. As such, it is not about memory allocation in anyway.
> > Before this change, the KV language sets the `text_size` to one of two values based on the value of `allow_stetch`. After this change the KV language sets `text_size` to the same value regardless of the value of `allow_stretch`.
> 
> I'm referring to the text above. That memory is only used for that code line, if you remove that code line, you should remove the other line as well, otherwise it will be left forgotten.

I am still not sure I am totally getting what you are saying but I think you are saying that `allow_stretch` is not used any where else which is not right - there is the method that implements the stretch functionality :

```
   def _update_tab_bar(self, *args):
        if self.parent.allow_stretch:
            # update width of the labels when it is needed
            width, tabs = self.scrollview.width, self.layout.children
            tabs_widths = [t.min_space for t in tabs if t.min_space]
            tabs_space = float(sum(tabs_widths))

etc ...
```

podraco:
So far, I can see, it's embedded with `self._trigger_update_tab_bar`  which only stores a result, rather a partial function...

podraco:
I believe that `_update_tab_bar`  is the one not needed.
Since I removed it and the widget works perfectly.
Also added some functions required to simplify the widget work with dynamic title labels.
here's an example.


https://user-images.githubusercontent.com/5509325/108430206-a0518780-7206-11eb-93ed-411c6c4a3e04.mp4


so far i can see, the function was doing the same as the line 520.
the line by itself is not of great help, but, if the code is added to the `on_text` event instead `on_tab_bar` the label updates it's size automatically.

runette:
> I believe that `_update_tab_bar` is the one not needed.
> Since I removed it and the widget works perfectly.
> Also added some functions required to simplify the widget work with dynamic title labels.
> here's an example.

The problem is (and this is not the subject of this PR ) that the `allow_stretch` functionality only has any effect when the total length is less than the length of the Tab Bar (that is not anything I have decided - that is the way it has always been). 

So - your test case is wrong and if you base on that your solution will be wrong.

For instance - this test case:

```
from kivy.lang import Builder
from kivy.uix.floatlayout import FloatLayout

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase

KV = '''
BoxLayout:
    orientation: "vertical"

    MDToolbar:
        title: "Example Tabs"

    MDTabs:
        id: tabs
        allow_stretch: True

<Tab>:

    MDLabel:
        text: "Content"
'''

class Tab(FloatLayout, MDTabsBase):
    '''Class implementing content for a tab.'''

class Test(MDApp):

    def build(self):

    def on_start(self):
        self.root.ids.tabs.add_widget(Tab(text="one"))
        self.root.ids.tabs.add_widget(Tab(text="two"))

Test().run()
```
This looks as follows in the current release:


https://user-images.githubusercontent.com/2239795/108485518-ea814a00-7294-11eb-961c-d94284c79b94.mp4

and like this of you disable `_update_tab_bar`:


https://user-images.githubusercontent.com/2239795/108485634-0ab10900-7295-11eb-81a8-ea6b16676d0b.mp4

HeaTTheatR:
@runette I think this pull request can be closed due to a recent merge - https://github.com/kivymd/KivyMD/pull/877

runette:
Unfortunately, I still have the problem in my App. I a trying to isolate the problem. 

I still can not see any reason why we need to set the `text_size` parameter.

runette:
This issue has NOT been solved in the latest release.

It can be reproduced in the latest version of Kitchen SInk.

NOTE - to reproduce this problem, the MDTab must be set to `allow_stretch=False`
NOTE ALSO - in the current version of Kitchen sink  - the `allow_stretch` attribute is applied to the wrong class - it must be applied to the MDTab class and not the MDTabBase class. If this is corrected - you get this

Screenshot 2021-02-20 at 00 18 37

podraco:
> This issue has NOT been solved in the latest release.
> 
> It can be reproduced in the latest version of Kitchen SInk.
> 
> NOTE - to reproduce this problem, the MDTab must be set to `allow_stretch=False`
> NOTE ALSO - in the current version of Kitchen sink - the `allow_stretch` attribute is applied to the wrong class - it must be applied to the MDTab class and not the MDTabBase class. If this is corrected - you get this
> 
> Screenshot 2021-02-20 at 00 18 37

that code does not correspond to the current master branch of kivymd, since this version sets the minimum width of a tab at `90 dp` when even when a size hint is used or not. And even so, the error keeps occurring, I must check what's happening, thanks for the heads-up :9 

This is the screen it gives me:
Captura de pantalla de 2021-02-19 19-37-27

podraco:
> Unfortunately, I still have the problem in my App. I a trying to isolate the problem.
> 
> I still can not see any reason why we need to set the `text_size` parameter.

it forces an update on the widget size.

runette:
> > Unfortunately, I still have the problem in my App. I a trying to isolate the problem.
> > I still can not see any reason why we need to set the `text_size` parameter.
> 
> it forces an update on the widget size.

Actually, it does not. The documentation on kivy is quite clear that`text_size` does not affect the widget size - it is only used to *limit* the size of the texture. Its purpose is to force the text to wrap or to truncate.

This is inherent in the line under discussion - since that sets `text_size` to `self.width`. If `text_size` forces an update to the widget size - that would be redundant.

So, in light of the current discussion - I could understand if `text_size` to 150dp - as the new proposed widest tab - but I do not see why it would ever be less than that since material.io specifically states that tabs should not wrap unless they need to.

See the following from the kivy docs:

Screenshot 2021-02-20 at 14 43 30


Screenshot 2021-02-20 at 14 44 00

runette:
> That code does not correspond to the current master branch of kivymd, since this version sets the minimum width of a tab at `90 dp` when even when a size hint is used or not. And even so, the error keeps occurring, I must check what's happening, thanks for the heads-up 

Actually - it Is the code in the current master.

And I think I know why we are seeing different things. I develop on Mac - and my app needs to run on Mac / Windows and Linux.

I have seen this before - although it is somewhat difficult to pin down and the documentation available is not clear. Sometimes - on a mac with a retina screen - `dp` seems to actually be half of what it is on a Windows screen.

I know that `dp` is supposed to avoid this problem - but it seems to have a bug with a mismatch between `dp` and retina. I admit that I have found better response with `sp` - which works better across platforms and would make more sense in a text label!

Note that this will also be a problem when you start to configure maximum tab label size - I believe they should be `sp` and not `dp`.

Note also - that this is not the cause or solution for this bug - those tab labels should not be wrapping at all! This is only the explanation for why our screens look different.

HeaTTheatR:
@runette Thanks for reviewing this issue in detail.

podraco:
That info is quite useful, and yet, concerning.
I didn't know of that behavior, since I can only use the app on Windows and Linux (and recently android)

Shouldn't that be reported to the kivy project?

podraco:
> > That code does not correspond to the current master branch of kivymd, since this version sets the minimum width of a tab at `90 dp` when even when a size hint is used or not. And even so, the error keeps occurring, I must check what's happening, thanks for the heads-up
> 
> Actually - it Is the code in the current master.
> 
> And I think I know why we are seeing different things. I develop on Mac - and my app needs to run on Mac / Windows and Linux.
> 
> I have seen this before - although it is somewhat difficult to pin down and the documentation available is not clear. Sometimes - on a mac with a retina screen - `dp` seems to actually be half of what it is on a Windows screen.
> 
> I know that `dp` is supposed to avoid this problem - but it seems to have a bug with a mismatch between `dp` and retina. I admit that I have found better response with `sp` - which works better across platforms and would make more sense in a text label!
> 
> Note that this will also be a problem when you start to configure maximum tab label size - I believe they should be `sp` and not `dp`.
> 
> Note also - that this is not the cause or solution for this bug - those tab labels should not be wrapping at all! This is only the explanation for why our screens look different.

the kivy documentation talks bout sp as :
Scale-independent Pixels - This is like the dp unit, but it is also scaled by the user’s font size preference. We recommend you use this unit when specifying font sizes, so the font size will be adjusted to both the screen density and the **user’s preference.**

wich means that maybe the issue is the preference of the user, if i remember right, Mac OS and other linux OS allows you to set a density/size for icons and fonts.

runette:
> the kivy documentation talks bout sp as :
> Scale-independent Pixels - This is like the dp unit, but it is also scaled by the user’s font size preference. We recommend you use this unit when specifying font sizes, so the font size will be adjusted to both the screen density and the **user’s preference.**
> 
> which means that maybe the issue is the preference of the user, if i remember right, Mac OS and other linux OS allows you to set a density/size for icons and fonts.

I agree that dp thing is very complicated - which is also why I will probably never report it to kivy and just work around it.

However - please, please do NOT get hung up or rabbit holed by the dp issue

The dp thing does NOT explain the core issue of this PR. 

The dp thing ONlY explains why your screen and my screen are equally wrong in slightly different ways. 

This issue is that the current version of MDTabs is setting `text-size` when it doe not need to and at the wrong time and that i causing my app to fail when it did not before.

That is the issue I urgently need to solve for my app to continue working without having to create a completely separate fork of the code.

runette:
> @runette Thanks for reviewing this issue in detail.

@HeaTTheatR - :) it is not altruism. I have invested some time in writing an app critical to my company in kivyMD.

podraco:
That explains why you know so well the behaviors and more practical stuff.

podraco:
ok, got some code done, do you want to test it before the next PR?

runette:
please

podraco:
Here is the fork, I'm reviewing the docs and adding some stuff.
https://github.com/podraco/KivyMD

runette:
I have some comments. I probably will not be able to send them until I finish work. Can I please ask that you do not push anything in the meantime?

podraco:
I just pushed a pull request, it's still in WIP

runette:
Thanks :)

That  makes it a lot easier to send review - I will do that on the PR.

runette:
This PR is closed by #887

Issue #770, opened 2021-01-13
MDTabsBar makes an incorrect assumption that x pos is 0


runette:
### Description of the Bug

I have found the base problem for the MDTabsBar autoscrolling error.

The autoscroll method asumes that x pos of the Tabs Bar is 0 - i.e. that it is the left most component or that it goes across the whole screen. This seems to be around this code (tabs.py lines 855 - 858):

```python
        bound_left = self.center_x
        bound_right = self.layout.width - bound_left
        dt = target.center_x - bound_left
        sx, sy = self.scrollview.convert_distance_to_scroll(dt, 0)
```
I think that line 855 should be

```python
bound_left = self.center_x - self.pos[0]
```

This seems to work but I am not sure enough of the corner cases of this code to actually make that a PR
### Code and Logs

This snippet puts the Tab in the right hand half of the view and autoscrolling does not work at all because the local center of individual tabs is always less than the left bound.

```python

from kivy.lang import Builder
from kivy.uix.floatlayout import FloatLayout
from kivy.properties import ObjectProperty

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase
from kivymd.icon_definitions import md_icons


KV = '''

BoxLayout:
    orientation: "horizontal"
    Widget:
    BoxLayout:
        orientation: "vertical"
        MDToolbar:
            title: "Example Tabs"
        MDTabs:
            id: tabs


<Tab>:

    MDIconButton:
        id: icon
        icon: root.icon
        user_font_size: "48sp"
        pos_hint: {"center_x": .5, "center_y": .5}
        
'''


class Tab(FloatLayout, MDTabsBase):
    '''Class implementing content for a tab.'''
    icon = ObjectProperty()


class Example(MDApp):
    icons = list(md_icons.keys())[15:30]

    def build(self):
        return Builder.load_string(KV)

    def on_start(self):
        for name_tab in self.icons:
            tab = Tab(text="This is " + name_tab, icon=name_tab)
            self.root.ids.tabs.add_widget(tab)



Example().run()

```

### Screenshots

Add images to explain us this bug. Paste urls here.

Remove this section if no images here

### Versions

* OS: mac
* Python: 3.8.6
* Kivy: 2.0.0
* KivyMD: Master

podraco:
MDTabsBar autoscrolling error?
is a reported issue?

runette:
#767

podraco:
[...]

PR #877, opened 2021-02-19
WIP - MDTabs: bugfixes and improvements

podraco:
### Description of Changes

- Fixed animation, issue #770

- Fixed tab size,  issue #874

- Fixed toggle behavior missing group release.

- Fixed Toggle behavior when the tab container is displayed on screen (the 
current label wasn't toggled down, and the group wasn't released)

- Simplify the work with icons, text and mixed labels.

- Added documentation for every new property.

- Fixed some minor issues.

- Kitchen sink demo was updated.

- Indicator canvas instruction fixed:
The prior method used the class instance instead of a group name inside  the canvas layers.
This method ensures that the instruction referred to is always the correct.

- Minor doc updates on event description.

- Indicator Animation simplified.

- This fix allows the indicator to change size every time theres a label 
size update.

- Removed some unused kv code

- Simplified Android Animation processing.

#### NEW:

1.- you can set the title label directly with text in the current 
font_style for buttons.

2.- You can now set the icon easier and edit it in a simplified way for 
dynamic title labels.

3.- Title label complies with Material IO spec definitions.

4.- You can set the title label mode for mixed labels (Modes available `"Lead"`, `"Top"`) inside each tab.

5.- You can now force a mode for every tabs label title.

6.- The scroll_view in which the title labels are stored now changes size depending on the mode selected. (only works for fixed mode. Done inside class MDTabs():)

7.- Now you can get the slides list directly from class MDTabs.

8.- the tabs label minimum size is set to 90 dpi as the Material IO 
requires.

9.- removes some trigger updates that weren't necessary. (duplicated the behavior form MDLabel, and its container.)

10.- The kitchen sink demo features the current simplified method to update the tab's label.

#### Before

https://user-images.githubusercontent.com/5509325/108486224-79756e80-7263-11eb-89e8-1af9ee47bcd5.mp4

#### After

https://user-images.githubusercontent.com/5509325/108486236-7c705f00-7263-11eb-8b20-993e48198265.mp4

### DIFF
--- a/demos/kitchen_sink/libs/baseclass/tabs.py
+++ b/demos/kitchen_sink/libs/baseclass/tabs.py
@@ -5,40 +5,52 @@ from kivymd.icon_definitions import md_icons
 
 
 class KitchenSinkTabs(Screen):
+    # Icons to be used to populate the tabs.
     list_name_icons = list(md_icons.keys())[0:15]
+    # lock the memory to avoid duplicity.
     tabs_created = False
 
     def on_enter(self):
+        # updates the screen afther the animation ends.
         if not self.tabs_created:
             for name_tab in self.list_name_icons:
                 self.ids.android_tabs.add_widget(
-                    Factory.KitchenSinkTabItem(text=name_tab)
+                    Factory.KitchenSinkTabItem(title=name_tab)
                 )
             self.tabs_created = True
 
     def switch_tabs_to_text(self, instance_android_tabs):
-        for instance_tab in instance_android_tabs.ids.scrollview.children[
-            0
-        ].children:
-            for k, v in md_icons.items():
-                if v == instance_tab.text:
-                    instance_android_tabs.ids.scrollview.children[
-                        0
-                    ].remove_widget(instance_tab)
-                    instance_android_tabs.add_widget(
-                        Factory.KitchenSinkTabItem(
-                            text=" ".join(k.split("-")).capitalize()
-                        )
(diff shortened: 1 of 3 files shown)


HeaTTheatR:
@podraco Good job!
As far as I understand, we don't need to merge these changes with the master branch yet?..

podraco:
I believe It can be merged, but I'm not sure if the docs are correct

HeaTTheatR:
@podraco 
Why aren't you testing the changes in the demo application?

```python
    def on_release(self):
        self.tab_bar.parent.dispatch("on_tab_switch", self.tab, self, self.text)
```

HeaTTheatR:
@podraco

https://user-images.githubusercontent.com/16930280/109657882-08915900-7b77-11eb-91ab-8b3deb199276.mov

HeaTTheatR:
@podraco If you have made changes to the Tabs class, you should test the functionality of the demo application completely.

podraco:
?????
I used the demo and the examples in the docs to see if it worked properly. In my end didn't show any error.

HeaTTheatR:
@podraco 
The demo application uses all the widgets in the library, and if we have made any changes, we must first test them in the demo application. On all counts. Since, for example, the element "Theming colors" uses tabs ...

podraco:
I didn't see any problem, sorry.
That's why the second pr fixed what was left broken

Issue #885, opened 2021-02-22
MDTabs - No Valid Icon was found.  No valid Title was found.  Icon   = ""    Title    = "" ->  kivymy.uix.tab.update_text() function 


spreadsomejelly:
### Description of the Bug
When using MDTabs -> getting _No Valid Icon was Found_ **ValueError** (line #758 in kivymy.uix.tab._update_text).  Tried setting title, icon, text independently, then together (several variations - to make sure I'm not losing my mind).  I tried using a separate KV file, and also putting a KV string inside the main.py file; however, no dice.  It seems as though the MDTabsBase class is instantiating without the text value, even though the parameter is set.  Thoughts?  

### Code and Logs

```python
import kivy # 96 dpi
#kivy.require('1.11.1')
from kivy.app import App
from kivymd.app import MDApp
from kivymd.uix.boxlayout import MDBoxLayout
from kivymd.uix.tab import MDTabsBase
from kivymd.uix.floatlayout import MDFloatLayout

kv = """
MDBoxLayout:
	MDTabs:
		Tab:
			#text: 'Work'
			#title: 'Now'
			icon: 'chevron-up'
		Tab:
			#text: 'Please'
			#title: 'Goddamnit'
			icon: 'chevron-down'
"""
class Tab(MDFloatLayout, MDTabsBase):

	def __init__(self, **kwargs):
		super(Tab, self).__init__(**kwargs)

class TestApp(MDApp):
	
	def build(self):
		pass # when using the KV File
                #Builder.load_string(kv)
               
if __name__ == '__main__':
	TestApp().run()

---------------------------------------------------------------------------------
#(KV File - same instructions)
# ROOT
MDBoxLayout:
	MDTabs:
		Tab:
			#text: 'Work'
			#title: 'Now'
			icon: 'chevron-up'
		Tab:
			#text: 'Please'
			#title: 'Goddamnit'
			icon: 'chevron-down'
"""
```
### ERROR/Trace
[WARNING] stderr:   File "C:\Users\spreadsomejelly\.virtualenvs\banana_pudding_pops\lib\site-packages\kivymd\uix\tab.py", line 732, in __init__
[WARNING] stderr:     self._update_text()
[WARNING] stderr:   File "C:\Users\spreadsomejelly\.virtualenvs\banana_pudding_pops\lib\site-packages\kivymd\uix\tab.py", line 758, in _update_text
[WARNING] stderr:     f"{self}: [UID] = [{self.uid}]:\n\t"
[...]

Issue #880, opened 2021-02-19
MDTabs: Unnecessary hard coded formating


runette:
### Description of the Bug

There are three areas where the new MDTabs hard codes formatting and styles. This is bad practice.

1. The `font_style` is hardcoded to 'Button'. This should be an OptionProperty defaulted to 'Button'
2. The `font_size` is hardcoded to `sp(14)` - this should be taken from the definition for the font_style that is currently chosen. As a side note - it should be taken from the instance of the font_definition currently loaded into the app theme_cls to retain the correct theming behavior.
3. When the title is set by the `title` attribute, it is hardcoded to be upper case. This is just wrong. If the app programmer wants upper case - they can easily set it themselves. This should NOT be hard coded into the framework and is NOT specified in material.io.

runette:
Actually - on line 758 not only is the title made upper case for a second time it is also hardcoded to be bold. material.io does not specify that the tab label should be bold.

podraco:
Last time I read it, The material IO defines that the Tabs should use the Button style, and bold. Let me check it.

podraco:
I was wrong I misunderstood the info, and applied the example spec definition to the tabs label.
What do you propose to be the current tab style?

podraco:
Even tho, the tabs are buttons, should we add a theme style pointer, so the end developer could set the style manually?

HeaTTheatR:
@podraco What do you mean by "button style"?

runette:
I suggest, or propose, the following:

1 No formatting should be hardcoded - it should all be a matter of configuration
2 create a font_style property and default to "Button" font_style but allow the app dev to change it
3 derive the formatting from the font style. As I said above - this needs to be derived from the `font_style` property of the current `theme_cls` so that theme changes to the app are correctly reflected.

The changes needed are not complicated and I can PR a suggestion quite quickly - I am just aware that this class is under development and I don't want to do it unilaterally.

podraco:
> @podraco What do you mean by "button style"?

label.font_style="Button", from the theme manager.

podraco:
> I suggest, or propose, the following:
> 
> 1 No formatting should be hardcoded - it should all be a matter of configuration
> 2 create a font_style property and default to "Button" font_style but allow the app dev to change it
> 3 derive the formatting from the font style. As I said above - this needs to be derived from the `font_style` property of the current `theme_cls` so that theme changes to the app are correctly reflected.
> 
[...]

Issue #878, opened 2021-02-19
MDTabs: allow_stretch not working


runette:
### Description

With the latest update to MDTabs - `allow_stretch` is not working. This is because you have removed the code that implemented it.

### 

Run this test application:

```
from kivy.lang import Builder
from kivy.uix.floatlayout import FloatLayout

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase

KV = '''
BoxLayout:
    orientation: "vertical"

    MDToolbar:
        title: "Example Tabs"

    MDTabs:
        id: tabs
        allow_stretch: True

<Tab>:

    MDLabel:
        text: "Content"
        

'''

class Tab(FloatLayout, MDTabsBase):
    '''Class implementing content for a tab.'''

class Test(MDApp):

    def build(self):
        return Builder.load_string(KV)

    def on_start(self):
        self.root.ids.tabs.add_widget(Tab(text="One"))
        self.root.ids.tabs.add_widget(Tab(text="Two"))


Test().run()
```

With the current version - you get this:

<img width="800" alt="Screenshot 2021-02-19 at 21 18 58" src="https://user-images.githubusercontent.com/2239795/108563120-dc651500-72f8-11eb-8727-40969636cdc5.png">

With the previous version - you get this :

https://user-images.githubusercontent.com/2239795/108563183-f43c9900-72f8-11eb-94ed-01c6c06c6ef1.mp4

* OS: macos
* Python: 3.9
* Kivy: 2.0.0
* KivyMD: Master Latest

podraco:
It is true, a bug, and yet wrong. The spec states that tabs can only be 150dp long, I'm working on it right now.

podraco:
![Captura de pantalla de 2021-02-20 01-29-18](https://user-images.githubusercontent.com/5509325/108587763-19da9a00-731b-11eb-9dc7-1dce9afac7f3.png)

so far, it's fixed.
Now, it updates the text_size to None, None, and then updates it again to lock the wrapping. I'm testing different configurations.
Added the scrollable padding when the tab navigation is scrollable.
Blocked the scroll when the width from GridLayout is lesser than the scrollview one.

podraco:
After all night testing and debugging i believe it will work now, ill open a Pr in the morning, cant read anymore XD

podraco:
as a quick resume of the changes
[...]

PR #887, opened 2021-02-23
MD Tab Update

podraco:
### Description of Changes
this is a fix for some problems that surged from the last pull request, as well add some improvements and upgrades to the class.

Possible fix for #770 
Possible fix for #874 
Possible fix for #878 
Possible fix for #880 
Possible fix for #885

This is a big change for the class, it has been proven with every example code provided by the documentation and kitchen sink.

This updates the documentation for the new implemented methods.

Between the fixes there are:
- simplified the `remove_widget ` function from `MDTabs`
- `remove_widget ` now removes binds created on `add_widget` for new tabs.
- max width of every tabt's title label is set to `"360dp"`
- both kv and python instance creation reviewed. (small differences that affected its behavior on new widgets)
- added exception raises if something is not filled correctly or found.
- Added more specific Exceptions to allow developers know what exactly went wrong.
- Added basic `MDCarousel` Documentation.
- `MDTabsBase.text` is now deprecated, still has backward compatibility, but the use `tab_label_text` is now enforced and the use of this property will show a message in the kivy logger.
- Fixed reference bug in which, if the tab with the reference was pressed, the selected tab would be missing its state.
- Tab_bar now behaves on the content's width. 
If the width of the tab's inside is bigger than the scroll view, the `tab_bar` will be set to scrollable, the padding of the layout will be set to `["52dp", 0]` and allow scroll, if the width is lesser or equal to the scrollview width, the scroll then is locked, and the canvas padding is set to `[0,0]`,  `ScrollView.do_scroll_x` is set to `False` in this scenario.
- Some duplicated code was removed
- Some duplicated movements were removed, since some steps were already done by the children widget or by another event.    
- Fixed some small recursions.

Adds:
- tab_label_text namespace will store the raw tab's title label text. Replaces `MDTabsBase.text` property.
-  Tabs.switch_to is now more flexible, you can toggle between text, Tab_objects and if it's a string, search by its icon, title or raw text.
- Any text parsed to the Title property will be set with upper(). (this is to maintain every tab title clear)
- added title_is_capital, if the value is True, it will update the self. Text to be all capitals.

### DIFF
--- a/kivymd/theming.py
+++ b/kivymd/theming.py
@@ -166,8 +166,8 @@ from kivy.utils import get_color_from_hex
 
 from kivymd import images_path
 from kivymd.color_definitions import colors, hue, palette
-from kivymd.material_resources import DEVICE_IOS, DEVICE_TYPE
 from kivymd.font_definitions import theme_font_styles
+from kivymd.material_resources import DEVICE_IOS, DEVICE_TYPE
 
 
 class ThemeManager(EventDispatcher):
@@ -1057,7 +1057,7 @@ class ThemeManager(EventDispatcher):
         self.colors = colors
         Clock.schedule_once(self.sync_theme_styles)
 
-    def sync_theme_styles(self,*args):
+    def sync_theme_styles(self, *args):
         # Syncs the values from self.font_styles to theme_font_styles
         # this will ensure continuity when someone registers a new font_style.
         for num, style in enumerate(theme_font_styles):
@@ -1065,7 +1065,6 @@ class ThemeManager(EventDispatcher):
                 theme_font_styles.pop(num)
         for style in self.font_styles.keys():
             theme_font_styles.append(style)
-
 
 
 class ThemableBehavior(EventDispatcher):
--- a/kivymd/theming.py
+++ b/kivymd/theming.py
@@ -167,6 +167,7 @@ from kivy.utils import get_color_from_hex
 from kivymd import images_path
 from kivymd.color_definitions import colors, hue, palette
 from kivymd.material_resources import DEVICE_IOS, DEVICE_TYPE
+from kivymd.font_definitions import theme_font_styles
 
 
 class ThemeManager(EventDispatcher):
@@ -1052,7 +1053,19 @@ class ThemeManager(EventDispatcher):
         Clock.schedule_once(lambda x: self.on_theme_style(0, self.theme_style))
         self._determine_device_orientation(None, Window.size)
         Window.bind(size=self._determine_device_orientation)
+        self.bind(font_styles=self.sync_theme_styles)
         self.colors = colors
+        Clock.schedule_once(self.sync_theme_styles)
+
+    def sync_theme_styles(self,*args):
+        # Syncs the values from self.font_styles to theme_font_styles
+        # this will ensure continuity when someone registers a new font_style.
+        for num, style in enumerate(theme_font_styles):
+            if style not in self.font_styles:
+                theme_font_styles.pop(num)
+        for style in self.font_styles.keys():
+            theme_font_styles.append(style)
+
 
 
 class ThemableBehavior(EventDispatcher):
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -708,8 +708,10 @@ class MDTabsBase(Widget):
     and defaults to `''`.
     """
 
-    text = StringProperty()
+    text = StringProperty(deprecated=True)
     """
+    This property is deprecated, use :attr:`tab_label_text` instead,
+
     This property is the actual title of the tab.
     use the property :attr:`icon` and :attr:`title` to set this property
     correctly.
@@ -718,6 +720,19 @@ class MDTabsBase(Widget):
     purposes.
 
     :attr:`text` is an :class:`~kivy.properties.StringProperty`
+    and defaults to `''`.
+    """
+
+    tab_label_text = StringProperty()
+    """
+    This property is the actual title's Label of the tab.
+    use the property :attr:`icon` and :attr:`title` to set this property
+    correctly.
+
+    This property is kept public for specific and backward compatibility
+    purposes.
+
+    :attr:`tab_label_text` is an :class:`~kivy.properties.StringProperty`
     and defaults to `''`.
     """
 
@@ -760,6 +775,8 @@ class MDTabsBase(Widget):
             icon=self._update_text,
             title=self._update_text,
             title_icon_mode=self._update_text,
+            text=self.update_label_text,
+            tab_label_text=self.update_label_text,
         )
         Clock.schedule_once(
             self._update_text
@@ -805,10 +822,13 @@ class MDTabsBase(Widget):
                 self.tab_label.height = dp(72)
         else:
             self.tab_label.height = dp(48)
-        self.on_text(None, None)
+        self.update_label_text(None, self.tab_label_text)
 
+    def update_label_text(self, widget, value):
+        self.tab_label.text = self.tab_label_text
+        
     def on_text(self, widget, text):
-        self.tab_label.text = self.text
+        self.tab_label_text = self.text
 
 
 class MDTabsMain(MDBoxLayout):
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -1321,7 +1321,9 @@ class MDTabs(ThemableBehavior, SpecificBackgroundColorBehavior, AnchorLayout):
         if not self.ids.layout.children:
             return
         if self.tab_hint_x is True:
-            self.fixed_tab_label_width = self.width//len(self.ids.layout.children)
+            self.fixed_tab_label_width = self.width // len(
+                self.ids.layout.children
+            )
             self.allow_stretch = False
         else:
             self.allow_stretch = True
@@ -1428,7 +1430,8 @@ class MDTabs(ThemableBehavior, SpecificBackgroundColorBehavior, AnchorLayout):
                     fixed_tab_label_width=widget.tab_label._update_text_size,
                     font_name=widget.tab_label.setter("font_name"),
                     text_color_active=widget.tab_label.setter(
-                        "text_color_active"),
+                        "text_color_active"
+                    ),
                     text_color_normal=widget.tab_label.setter(
                         "text_color_normal"
                     ),
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -566,9 +566,6 @@ Builder.load_string(
             anim_move_duration: root.anim_duration
             on_index: root.on_carousel_index(*args)
             on__offset: tab_bar.android_animation(*args)
-            on_slides:
-                self.index = root.default_tab
-                root.on_carousel_index(self, 0)
 
     MDTabsBar:
         id: tab_bar
--- a/kivymd/uix/tab.py
+++ b/kivymd/uix/tab.py
@@ -1174,7 +1174,12 @@ class MDTabs(ThemableBehavior, SpecificBackgroundColorBehavior, AnchorLayout):
 
     allow_stretch = BooleanProperty(True)
     """
-    If False - tabs will not stretch to full screen.
+    If `True`, The tab will update dynamically to it's content width, and wrap
+    any text if the widget is bigger than `"360dp"`.
+
+    If `False`, the tab won't update to it's maximum texture width.
+    this means that the `fixed_tab_label_width` will be used as the label
+    width.
 
     :attr:`allow_stretch` is an :class:`~kivy.properties.BooleanProperty`
     and defaults to `True`.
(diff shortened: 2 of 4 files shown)


podraco:
i have done multiple tests with different codes that where available on the issues reports and on the documentations and demos. i believe this changes works and fulfill the normal behavior of any kivyMD and comply with the Material IO standard.

If there's something that needs to be fixed, ill work on it as soon as possible as I can.

podraco:
I'm waiting response from a collaborator, please don't approve merge yet.

runette (commented):
As a general comment - I still have the problem that we are fundamentally changing the effect that `allow_stretch` has. This means that ALL previous apps that used `allow_stretch` MUST be changed and this does not even seem to be documented.

BEFORE (i.e. before about 2 weeks ago):

`allow_stretch=True` created the behavior described in material.io as "Fixed Tabs" - i.e. if possible it stretched the Tabs wider than the label text so that they filled the TabBar.

AFTER (i.e. now) :

`allow_stretch=True` means that the Tab will expand to the width of the text (i.e. pretty much what `allow_stretch = False` meant before.

I find it unfortunate and unnecessary to fundamentally reverse the meaning of an existing API element like this. It would be much better to deprecate the old behavior and create a new attribute to trigger the new behavior.

kivymd/uix/tab.py
```diff
@@ -720,40 +788,81 @@ class MDTabsBase(Widget):
     and defaults to `None`.
     """
 
+    def _get_label_font_style(self):
+        if self.tab_label:
+            return self.tab_label.font_style
+
+    def _set_label_font_style(self, value):
+        if self.tab_label:
+            if value in theme_font_styles:
```
runette, line 805: This shold be compared to the actual and current font_style and not the imported one. Since TabLabel inherits from MDLabel it also inherits from ThemeableBehavior so this should be 

`if value in self.tab_label.theme_cls.font_styles`

kivymd/uix/tab.py
```diff
@@ -481,7 +520,7 @@ def switch_tab(self):
 from kivy.uix.widget import Widget
 from kivy.utils import boundary
 
-from kivymd.font_definitions import fonts
+from kivymd.font_definitions import fonts, theme_font_styles
```
runette, line 523: As a general comment - components should use the ThemableBehavior to get the current font_definitions and not import them - since they may have been changed by the application designer in the App.

kivymd/uix/tab.py
```diff
@@ -720,40 +788,81 @@ class MDTabsBase(Widget):
...
+        Clock.schedule_once(
+            self._update_text
+        )  # This will ensure the text is correct
 
     def _update_text(self, *args):
-        x = False
+        # ensures that the title is in capital letters.
         if self.title:
-            self.title = self.title.upper()
+            if self.title != self.title.upper():
```
runette, line 851: PLEASE - do not hard code the label title to Upper case.

This is a deal-breaker. There is NO specification in material.io that says that Tab labels should be all caps.

While you might think it is good practice and I might agree with you - this MUST be a matter for the app designer.

kivymd/uix/tab.py
```diff
@@ -1079,12 +1172,25 @@ class MDTabs(ThemableBehavior, SpecificBackgroundColorBehavior, AnchorLayout):
...
+
+    If `False`, the tab won't update to it's maximum texture width.
+    this means that the `fixed_tab_label_width` will be used as the label
+    width. this will wrap any text inside to fit the fixed value.
 
     :attr:`allow_stretch` is an :class:`~kivy.properties.BooleanProperty`
     and defaults to `True`.
     """
 
+    fixed_tab_label_width = NumericProperty("140dp")
```
runette, line 1203: See the general comment about the change in the meaning of `allow_stetch`

In addition - although this mechanism works and is broadly in line with material.io - I do miss the functionality to ask kivyMD to expand the Tabs to fill the screen. It was useful functionality and was also in line with material.io

runette (changes requested):
Starting to look good.

I do see some fairly fundamental bugs - as listed below

runette:
General Comment :

When I run the new code in Kitchen Sink - there is a gap to the left of the first tab that I don't remember being there before. I don't have the time today to do a side-by-side analysis to confirm that. It does not look like a phantom tab because it cannot be selected.

Screenshot 2021-02-23 at 11 15 05

runette:
General Comment :

When I run this new tab.py in my existing (released) application without other changes - I get this error message.

Throwing an error on legacy code that was working before could be a significant problem for the existing users of kivyMD.
```
builtins.ValueError: switch_tab:
name_tab not found in the tab list
search_by = 'text'
name_tab = 'Basic Details'
```

HeaTTheatR:
@runette  Didn't know the labels were so different on `Mac OS` :)

Без имени

runette:
> @runette Didn't know the labels were so different on `Mac OS` :)

I am using a custom font - mostly to test the most extreme case but also because my app uses the corporate brand font.

kivymd/uix/tab.py
podraco: The documentation stated that was the proper use to register and read a new font. Let me test it in a different demo to see if that works, if not, a new issue must be opened

kivymd/uix/tab.py
podraco: Well, the material io requieres that all labels must be clear, and. As far as I know, common practice to this is to use capital letters, since they have the most serif and use a bigger space to be drawn. But there's no problem, it can be removed

kivymd/uix/tab.py
podraco: It is a useful attribute, but, that shouldn't be it's name, since hat you're requiring, it's size_hint_x, from the grid layout that contains all tabs, implementing something to do the same in cpython to normal python is a really bad practice, but, i could forward the size hint of the tab bar to the MDTabsBase. So you could still use something like.
```py
tab_hint_x=1
tab_hint_x=None
```

kivymd/uix/tab.py
runette: As is usual, the problem is actually in how you state the question!

In this code - you should not be trying to register or read a *new* font. You are trying to read the fonts and actually font_styles that *have been* registered with the app.

If you use the current code - you will *only* get the standard fonts provided out of the box. If the app developer has followed the procedure as documented to add a new font to their app - it will be ignored by your code. This is obviously a bug in this code. I do not see the need for a new issue.

podraco:
That gap is required for scrollable tabs, it's in the specs and it's supposed to be `"52dp"`
If the width of every children in the container is lessed than the scroll view width, the do_scroll_x is False, and the padding =[0,0], this behavior updates every time the width is updated or a tab is added/removed/updated.

podraco:
> General Comment :
> 
> When I run this new tab.py in my existing (released) application without other changes - I get this error message.
> 
> Throwing an error on legacy code that was working before could be a significant problem for the existing users of kivyMD.
> 
> ```
> builtins.ValueError: switch_tab:
> name_tab not found in the tab list
> search_by = 'text'
> name_tab = 'Basic Details'
> ```

That was my error, forgot to simplify the text than to use the property name as the default value

podraco:
> General Comment :
> 
> When I run the new code in Kitchen Sink - there is a gap to the left of the first tab that I don't remember being there before. I don't have the time today to do a side-by-side analysis to confirm that. It does not look like a phantom tab because it cannot be selected.
> 
> Screenshot 2021-02-23 at 11 15 05

its because of this.

Captura de pantalla de 2021-02-23 15-14-06

Captura de pantalla de 2021-02-23 15-57-38

kivymd/uix/tab.py
podraco: Well, as far as i know, when a dic is imported. The `__import__` method only creates a link to the pointer, and add the value to the global environment rather to contain it in the file/lib. Because of this, you could alter the value in any file and every change should be shared. Yet, I'm not certain if the event dispatcher would call any event to it.

Also, as far as I know, comparing a value to a list were always faster than to a dict.

Captura de pantalla de 2021-02-23 17-38-17
Captura de pantalla de 2021-02-23 17-37-22
Captura de pantalla de 2021-02-23 17-37-33
Captura de pantalla de 2021-02-23 17-39-16

kivymd/uix/tab.py
podraco: yet, i just changed it to be optional.

kivymd/uix/tab.py
podraco: The code only uses what's stored in memory.

kivymd/uix/tab.py
podraco: The last commit ensures parity between the pointer and the theme_cls object.

podraco:
So, these are the latest changes, fix some recursion calls by events, and adds what I believe was the last needed feature.
I'll wait until it can be reviewed.

I suggest that we merge this until everyone involved approves this PR.

kivymd/uix/tab.py
runette: That is misdirection. It is all stored in memory. The question is whether you are using the correct values in the correct memory.

kivymd/uix/tab.py
runette: To me, you are adding complexity on top of complexity and adding a lot of new code and possible error modes for the relatively small performance gain for getting the list of values from a list rather than dict items. 

The class `ThemeManager` HAS an existing attribute that holds the current font styles. Why not just always drive from the current value of that attribute rather than require another value to be kept up to date?

If I was technical lead on this project I would say this solution adds to the error vectors of the project and makes the project less maintainable - and indeed would probably not give this change "readability' approval. But I think this will work.

kivymd/theming.py
```diff
@@ -1052,7 +1053,18 @@ def __init__(self, **kwargs):
...
+        Clock.schedule_once(self.sync_theme_styles)
+
+    def sync_theme_styles(self, *args):
+        # Syncs the values from self.font_styles to theme_font_styles
+        # this will ensure continuity when someone registers a new font_style.
+        for num, style in enumerate(theme_font_styles):
+            if style not in self.font_styles:
+                theme_font_styles.pop(num)
+        for style in self.font_styles.keys():
+            theme_font_styles.append(style)
```
runette, line 1067: See comment below.

To me - this looks like a complex solution to fix a problem that does not really exist.

kivymd/uix/tab.py
runette: 1. It is not a matter of whether you think the name is a good name or whether I think it is a good name - IT IS THE EXISTING API. That is what existing application - some of them things that people make money out of - have included in their applications. Not because they are stupid or bad programmers but because that is what the previous specification WAS.

2. I did not suggest anything about implementation - so I am not sure why you seem to think that I was proposing to change something from cython to python? 

3. I am not sure that I understand the proposed solution - not enough info - but if it involves adding NEW complexity to an already complex API then probably not.

I think what I would really would have liked is:

- `allow_stretch` doing exactly what it did before (because :: legacy) and deprecated (because :: not really MD and badly named), and

- A "tab mode" attribute that has values of  "Fixed", "Scrollable" and "Fixed-length" for the various modes in material.io (or something similar to this).

kivymd/uix/tab.py
runette: > As far as I know, common practice to this is to use capital letters, since they have the most serif and use a bigger space to be drawn

Thanks for making it optional.

Hard coding "good practice" is really not a good practice. And I do not necessarily agree with you. I use san-serif fonts (and Roboto is also san-serif) so that the question of the size of the serif has no relevance.

And if you look at from the point of view of psychology - all-caps words are harder for humans to read because we actually read by pattern recognition on the word and not by reading the letters.

runette:
I have no further technical issues with this PR - it is working for me both in my app an in kitchen garden.

My only remaining concern is the breaking change in the API (i.e. the change in meaning and usaga of `allow_stretch`).

I will leave that to @HeaTTheatR to decide.

runette:
BTW - just a procedural tip that you probably know already - if instead of 

"Possible fix for #770"
...

you had 

"fixes #770"
...

then merging the PR would automatically close the issue.

see https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

syleishere:
Just tested this again, allow_stretch still won't work without tab_hint_x: True

podraco:
Allow stretch is by defect `True`.
Allow stretch will allow the tab to stretch to the maximum size of the tab by the label content before wrapping it to 2 lines.
if `allow_stretch` is `False`, the tab won't change to the text texture, instead, it will use the fixed width property to set the tab's label width.

Meanwhile, `Tab_hint_x` will work outside the tab's texture size, and allow the tab to use a size hint based on the MDTabs widget width.

kivymd/theming.py
podraco: Both namespaces should keep the same keys, values, otherwise, why should it exists as separated elements?

PR #914, opened 2021-03-14
Datatable bugfixes and features

kevin-schumann:
### Description of Changes

- Page pagination now displays the correct data after sorting a column
- Checkboxes will now persist even after sorting a column
- - **This requires a change of the user-provided sort function! See sample code below.**
- (Un)checking all rows now (un)checks every row on every page
- Datatable no longer crashes when using the page pagination after sorting a column
- on_check_press no longer gets called when changing pages

### Known issues
- When (un)checking all rows only the displayed rows receive the on_check_press event

### Sample code

```python
from kivy.metrics import dp
from kivy.uix.anchorlayout import AnchorLayout

from kivymd.app import MDApp
from kivymd.uix.datatables import MDDataTable


class Example(MDApp):
    def build(self):
        layout = AnchorLayout()
        self.data_tables = MDDataTable(
            size_hint=(0.7, 0.6),
            use_pagination=True,
            check=True,
            column_data=[
                ("No.", dp(30)),
                ("Status", dp(30)),
                ("Signal Name", dp(60), self.sort_on_signal),
                ("Severity", dp(30)),
                ("Stage", dp(30)),
                ("Schedule", dp(30)),
                ("Team Lead", dp(30))
            ],
            row_data=[
                (1,) * 7,
                (2,) * 7,
                (3,) * 7,
                (4,) * 7,
                (5,) * 7,
                (6,) * 7,
                (7,) * 7,
                (8,) * 7,
                (9,) * 7,

            ],
            sorted_on="Schedule",
            sorted_order="ASC",
            elevation=2
        )

        self.data_tables.bind(on_check_press=self.on_check_press)
        layout.add_widget(self.data_tables)
        return layout

    def on_check_press(self, instance_table, current_row):
        '''Called when the check box in the table row is checked.'''

        print(instance_table, current_row)

    def sort_on_signal(self, data):
        ''' 
        Returns a
             - Tuple of indices (the permutation to sort the data in ascending order)
             - Tuple of the ascending ordered data
        '''
        return zip(*sorted(enumerate(data), key=lambda l: int(l[1][3])))

Example().run()
```

### DIFF
--- a/kivymd/uix/datatables.py
+++ b/kivymd/uix/datatables.py
@@ -22,6 +22,9 @@ Components/DataTables
 
 __all__ = ("MDDataTable",)
 
+from collections import defaultdict
+from math import ceil
+
 from kivy.clock import Clock
 from kivy.lang import Builder
 from kivy.metrics import dp
@@ -50,8 +53,6 @@ from kivymd.uix.boxlayout import MDBoxLayout
 from kivymd.uix.button import MDIconButton
 from kivymd.uix.menu import MDDropdownMenu
 from kivymd.uix.tooltip import MDTooltip
-from collections import defaultdict
-from math import ceil
 
 Builder.load_string(
     """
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@kevin-schumann 
At the time when I received a notification about your PR, I was making exactly the same corrections.
Good job!

javixav:
Hi kivy team, it might be not the right place to say this but the sort_on method doesnt work for the first column ^^, but works for all other columns.
#kivy 2.2.1
#kivymd 1.2.0
def sort_on_title(self, data):
        return zip(*sorted(enumerate(data), key=lambda l: l[1][0]))
The arrow does not show like in other columns

Okk I just created a new ticket in isssue kivymd page [here ](https://github.com/kivymd/KivyMD/issues/1819)

PR #925, opened 2021-03-21
Fix same icon MDFloatingActionButtonSpeedDial

VICTORVICKIE:
### Description of Changes
This fix allows user to use same icon for multiple speed dials in MDFloatingActionButtonSpeedDial
but data need to be changed as follows, 
{'icon_name': 'name to be displayed'}
```
data = {
            'Python':'plus' ,
            'PHP':'plus' ,
            'C++':'plus' ,
        }
```

### Screenshots
Screenshot_1

### DIFF
--- a/kivymd/uix/button.py
+++ b/kivymd/uix/button.py
@@ -349,9 +349,9 @@ MDFloatingActionButtonSpeedDial
 
     class Example(MDApp):
         data = {
-            'language-python': 'Python',
-            'language-php': 'PHP',
-            'language-cpp': 'C++',
+            'Python': 'language-python',
+            'PHP': 'language-php',
+            'C++': 'language-cpp',
         }
 
         def build(self):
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@VICTORVICKIE 
Thanks for PR!
The following data structure is not correct from the point of view of the Python programming language:

```python
data = {
            'plus': 'Python',
            'plus': 'PHP',
            'plus': 'C++',
        }
```
And then, I'm not sure that we should implement such a possibility as using two identical icons for a stack of buttons.

VICTORVICKIE:
Two identical icons may give app devs more access,
and I've commented wrong data dict structure
```
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDScreen:

    MDFloatingActionButtonSpeedDial:
        data: app.data
        root_button_anim: True
'''


class Example(MDApp):
    data = {
        'Python':'plus' ,
        'PHP':'plus' ,
        'C++':'plus' ,
    }

    def build(self):


Example().run()
```

HeaTTheatR:
@VICTORVICKIE 
Have you checked these changes on the Kitchen demo application?

VICTORVICKIE:
Checked and i ll send another updated PR

PR #934, opened 2021-03-24
Remove redundant defaults for MDSpinner

drexfall:
MDSpinner object takes in a ColorProperty as 'color' but gets immediately overridden by the 'primary_color' of the application. The solution proposed allows None value for 'color' and hence overrides it only if it is a None value. Also, any changes in app's primary color will be reflected only if the original color was set as the primary color. The canvas colors are also changed accordingly.

### DIFF
--- a/kivymd/uix/spinner.py
+++ b/kivymd/uix/spinner.py
@@ -133,7 +133,7 @@ Builder.load_string(
             origin: self.center
     canvas:
         Color:
-            rgba: self.color
+            rgba: self.color if self.color else self.theme_cls.primary_color
             a: self._alpha
         SmoothLine:
             circle: self.center_x, self.center_y, self.width / 2, \
@@ -196,7 +196,7 @@ class MDSpinner(ThemableBehavior, Widget):
     and defaults to `True`.
     """
 
-    color = ColorProperty([0, 0, 0, 0])
+    color = ColorProperty(None, allownone = True)
     """
     Spinner color.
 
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@shreyash
Thanks for the PR!
Have you tested these changes on a demo application? Have you tried dynamically changing the color palette and style of the app (dark and light theme)?

drexfall:
Yeah, I've checked. I've actually created a demo application just to understand this widget and I've attached the file link too.
[Demo](https://github.com/shreyash/demos/blob/main/kivy_spinner.py)

HeaTTheatR:
@shreyash I am asking this question because your changes break the MDSpinner widget.

HeaTTheatR:
@shreyash Try to dynamically change the color palette of the application and you will see what I mean.

kivymd/uix/spinner.py
```diff
@@ -220,14 +220,20 @@ class MDSpinner(ThemableBehavior, Widget):
...
+        if self.color == self.theme_cls.primary_color
+            self.theme_cls.bind(primary_color=self._update_color)
+            
         self._alpha_anim_in = Animation(_alpha=1, duration=0.8, t="out_quad")
         self._alpha_anim_out = Animation(_alpha=0, duration=0.3, t="out_quad")
         self._alpha_anim_out.bind(
             on_complete=self._reset,
             on_progress=self._on_determinate_progress,
         )
-        self.theme_cls.bind(primary_color=self._update_color)
```
HeaTTheatR, line 230: This binding updated the spinner palette when changing the application's color theme.

HeaTTheatR (changes requested):

kivymd/uix/spinner.py
drexfall: https://user-images.githubusercontent.com/67905454/112613336-fd63dd00-8e45-11eb-8227-162c610defc0.mp4

The binding is still intact for when the user mentions the color to be self.theme_cls.primary_color. If there is no color specified, the default primary color is taken and the binding still remains intact.

The problem which was, as you can see with the MDRaisedButtons in  the video attached, is that the binding forced an update regardless of the original value, leading to very obvious inconsistencies.

drexfall: [The code for above video](https://github.com/shreyash/demos/blob/main/kivy_spinner_palette.py)
The buttons update the entire app's palette and the spinner changes accordingly.

HeaTTheatR: @shreyash

https://user-images.githubusercontent.com/16930280/112614814-c46e3d00-8e32-11eb-996c-16d7aae0cdd3.mov

drexfall: @HeaTTheatR 
Can you check if it's not already broken? I cloned the repository, again and I'm having the same error as you.

PR #935, opened 2021-03-24
Cursor moving outside the box when multiline=True fixed.

Jonypr-code:
I modified textfield.py to solve the problem of the cursor moving out of the box. This image reports the bug that previously existed.
unknown
You can reproduce this issue by setting a max height and then add some lines to reach the max_height, and then scrowlup, and you will see the cursor moving outside the box.
This is a simple example to repruduce the issue:

```
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDScreen
    MDTextField:
        size_hint_x: .5
        hint_text: "multiline=True"
        max_height: "200dp"
        mode: "fill"
        fill_color: 0, 0, 0, .4
        multiline: True
'''

class Example(MDApp):
    def build(self):


Example().run()
```
With the latest commit, the three modes of textinput are fixed.

### DIFF
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -527,7 +527,6 @@ Builder.load_string(
         Rectangle:
             pos: (int(x) for x in self.cursor_pos)
             size: 1, -self.line_height
- 
         # Hint text.
         Color:
             rgba: self._current_hint_text_color if not self.current_hint_text_color else self.current_hint_text_color
--- a/kivymd/uix/textfield.py
+++ b/kivymd/uix/textfield.py
@@ -523,7 +523,7 @@ Builder.load_string(
         Color:
             rgba:
                 (self._current_line_color if self.focus and not \
-                self._cursor_blink and self.cursor_pos[1] < self.y + self.height - 20 and self.cursor_pos[1] > self.y + 20 else (0, 0, 0, 0))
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@Jonypr-code 
Thanks for the PR!
Please try to make more detailed commits: what exactly did you change, why.
`Update textfield.py` is not a very good commit since it doesn't contains any information.

Jonypr-code:
Yes i agree, im very proud of myself helping for helping this library, i will change the name of the pr, and i will change other thing in the code, you will see if it has `+20` it dont works with other modes, only with fill mode, the fix is just replacing to `+ 10`.

HeaTTheatR:
@Jonypr-code 
Please try to make more detailed commits: what exactly did you change, why.
`Update textfield.py` is not a very good commit since it doesn't contains any information.

HeaTTheatR:
@Jonypr-code Developers who, for example, will read the history of commits after seeing the text of your commit, will not understand anything.

maleksware:
The changes you applied are great, but pay attention to linter requirements. This is the only problem.

Jonypr-code:
Yes, i understand it, but i saw that the problem is with an unused import in a file that i didnt change nothing, the filemanager.py. I just changed the textfield.py.The other commits worked fine, i dont know why it changed to this last commit.

PR #940, opened 2021-03-29
Added reload method to FitImage class

Neizvestnyj:
Added a standard function for images - `reload`

Function reload image from disk. This facilitates re-loading of images from disk in case the image content changes.
If we change the image to another one with the same name after loading it, we will need to clear the cache, and this will be done by `reload`
```python
from kivymd.app import MDApp

KV = '''
Screen:
    FitImage:
        id: image
        source: 'Image.jpg'
        
    FloatLayout:
        MDIconButton:
            icon: 'reload'
            pos: root.width - self.width, root.height - self.height
            theme_text_color: "Custom"
            text_color: app.theme_cls.primary_color
            on_release: image.container.image.reload()
'''


class TestApp(MDApp):
    def build(self):


TestApp().run()

```

https://user-images.githubusercontent.com/40869738/112895474-d88d9500-90e5-11eb-9c99-eb096d0e401a.mov

[kivy image reload](https://kivy.org/doc/stable/api-kivy.uix.image.html?highlight=asyncimage#kivy.uix.image.Image.reload)

### DIFF
--- a/kivymd/utils/fitimage.py
+++ b/kivymd/utils/fitimage.py
@@ -122,6 +122,9 @@ class FitImage(BoxLayout):
         self.bind(source=self.container.setter("source"))
         self.add_widget(self.container)
 
+    def reload(self):
+        self.container.image.reload()
+
 
 class Container(Widget):
     source = ObjectProperty()


podraco:
I don't see any problem with this, good job.

HeaTTheatR:
> I don't see any problem with this, good job.

I’m not saying that it’s bad. I say that the "Updated FitImage" commit is not informative and should be changed to a more detailed one.

podraco:
how about
`FitImage - reload method added`
also, remember to update it's docs

HeaTTheatR:
> how about
> `FitImage - reload method added`
> also, remember to update it's docs

`Added `reload` method to `FitImage` class`

Neizvestnyj:
@HeaTTheatR Should I create a new commit and then a pull request?

HeaTTheatR:
> @HeaTTheatR Should I create a new commit and then a pull request?

Most likely you should make a new PR.

PR #955, opened 2021-04-11
Fixes #954 and Added auto_dismiss for FloatingActionButtonSpeedDial

### LINKED ISSUE #954: MDFloatingActionButtonSpeedDial data = {'nam': 'icon'} KeyError: icon when hint_animation=True
### Description of the Bug
when use MDFloatingActionButtonSpeedDial with hint_animation=True
clicking on Button to open items then when ever moving mouse .. error appear .
Error at [ kivymd\uix\button.py", line 1819 ]
  if self.data[instance.icon] == widget.text:
  KeyError: 'language-python'

it works fine for standard ver 0.104.1
i updated to latest ver to use new items like Live Editing Preview and some new widgets. 
thank you for hard work. i hope if i could help one day .. but still exploring the code .
### Code and Logs

```python
from kivymd.app import MDApp
from kivy.lang import Builder

kv = """

MDFloatingActionButtonSpeedDial:
    data: app.data
    hint_animation: True
    callback: lambda x: print(x.icon)
"""

class Test(MDApp):
    data = {'python icon!':'language-python'}
    def build(self):
        root = Builder.load_string(kv)
        return root

Test().run()

```

### Screenshots
Screenshot 2021-04-10 221204

### Versions

* OS: Windows 10 
* Python:  [INFO   ] [Python      ] v3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]
* Kivy:  [INFO   ] [Kivy        ] v2.0.0
* KivyMD:  [INFO   ] [KivyMD      ] 0.104.2.dev0,

VICTORVICKIE:
### Fixes #954 

1. Fixed by adding a reference name
2. Added auto_dismiss for FloatingActionButtonSpeedDial

- [x] Checked with KitchenSink

### DIFF
--- a/kivymd/uix/button.py
+++ b/kivymd/uix/button.py
@@ -413,6 +413,14 @@ You can set your color values ​​for background, text of buttons etc:
 .. seealso::
 
     `See full example <https://github.com/kivymd/KivyMD/wiki/Components-Button>`_
+
+You can use various types of dismissal on the stack:
+
+.. code-block:: kv
+
+    MDFloatingActionButtonSpeedDial:
+        auto_dismiss: True              #dismisses when clicked out of widget
+        dismiss_on_option_touch: True   #dismisses even when clicked on options
 """
 
 __all__ = (
@@ -1634,7 +1642,7 @@ class MDFloatingActionButtonSpeedDial(ThemableBehavior, FloatLayout):
     .. code-block:: python
 
         {
-            'name-icon': 'Text label',
+            'Text label': 'name-icon',
             ...,
             ...,
         }
@@ -1918,7 +1926,7 @@ class MDFloatingActionButtonSpeedDial(ThemableBehavior, FloatLayout):
         for name, name_icon in value.items():
             bottom_button = MDFloatingBottomButton(
                 icon=name_icon,
-                ref_name=name
+                ref_name=name,
                 on_enter=self.on_enter,
                 on_leave=self.on_leave,
                 opacity=0,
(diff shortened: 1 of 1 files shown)


kivymd/uix/button.py
```diff
@@ -1793,6 +1793,22 @@ def callback(self, instance):
...
+    """
+
+    dismiss_on_option_touch = BooleanProperty(False)
+    """
+    Whether to dismiss stack when clicked on one of the options.
+
+    :attr:`dismiss_on_option_touch` is a :class:`~kivy.properties.BooleanProperty`
+    and defaults to `False`.
+    """
+
```
HeaTTheatR: It seems to me that the `dismiss_on_option_touch` parameter is superfluous. It is enough to specify the `auto_dismiss` parameter.

kivymd/uix/button.py
```diff
@@ -1564,7 +1564,7 @@ class BaseFloatingLabel(
 
 
 class MDFloatingBottomButton(BaseFloatingBottomButton):
-    pass
+    ref_name = StringProperty()
```
HeaTTheatR: This is not a very good name for a parameter. This name does not reflect the essence.

VICTORVICKIE: its is not a parameter its just to reference with code itself

VICTORVICKIE: anyways i ll change to text

kivymd/uix/button.py
VICTORVICKIE: ok i'll remove redundant feature alone

Issue #995, opened 2021-05-25
TypeError: Cannot convert kivy.properties.NumericPropertyStorage to kivy.properties.BoundedNumericPropertyStorage


olek5andr:
Python == 3.7.6
kivy == 2.1.0.dev0
KivyMD == 0.104.2.dev0

App crashes after updating from kivy == 2.0.0 to kivy == 2.1.0.dev0.
On kivy==2.0.0 the app runs PERFECTLY on mac, but lags extensively on an Android device. That's the main reason behind updating to kivy master.

The Error message:

```
kivy.lang.builder.BuilderException: Parser: File "<inline>", line 16:
 ...
      14:    size_hint: None, None
      15:    height: dp(20) + lbl_txt.texture_size[1]
 >>   16:    width: lbl_txt.texture_size[0] + dp(24)
      17:
      18:    MDLabel:
 ...
 TypeError: Cannot convert kivy.properties.NumericPropertyStorage to kivy.properties.BoundedNumericPropertyStorage
```

PS. I thought this was an issue with kivy, but I was told it's actually caused by KivyMD - https://github.com/kivy/kivy/issues/7398#issuecomment-800550183

Many thanks in advance!

HeaTTheatR:
@olek5andr For now, you can use `Kivy version 2.0.0`. Changes have not yet been made to `KivyMD` that were made for `kivy.properties` in version `2.0.0`

quitegreensky:
@HeaTTheatR I think we can close this issue? fixed in #996

HeaTTheatR:
@quitegreensky Not yet, because your recent changes do not completely solve the problem. The width of the buttons is cut.

PR #996, opened 2021-05-28
Fixed BoundedNumericProperty error in BaseRectangularButton #995

quitegreensky:
### Description of Changes
Will fix `TypeError: Cannot convert kivy.properties.BoundedNumericPropertyStorage to kivy.properties.NumericPropertyStorage` when using kivy `v2.1.0.dev0` . 

The idea is to check the `width` value of  `BaseRectangularButton` inside `on_width` event and then resize it if the value is less than `88`. So we don't need to `BoundedNumericProperty` to check the minimum value.

### DIFF
--- a/kivymd/uix/button.py
+++ b/kivymd/uix/button.py
@@ -445,7 +445,6 @@ from kivy.lang import Builder
 from kivy.metrics import dp
 from kivy.properties import (
     BooleanProperty,
-    BoundedNumericProperty,
     ColorProperty,
     DictProperty,
     NumericProperty,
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@quitegreensky 
Yes it is possible. But the BoundedNumericProperty property is also used in the following modules:

```
uix/tooltip.py
uix/tab.py
uix/behaviors/elevation.py
uix/behaviors/backgroundcolor_behavior.py
```

quitegreensky:
@HeaTTheatR Since i noticed we don't have problem with using `BoundedNumericProperty` inside those modules. The issue occurs only after overriding `width` property. Maybe i`m missing something ?

HeaTTheatR:
@quitegreensky If you have the `Kivy` master branch installed, you can test these modules and check if an error occurs related to the `BoundedNumericProperty` property...

quitegreensky:
@HeaTTheatR Yeah i have kivy from master installed but getting no error from those modules

PR #997, opened 2021-05-29
Double and triple touch validation moved to the `delete_clock` function of the `on_touch_down` event

Neizvestnyj:
Double and triple click checks must occur when the `on_touch_down` method is triggered, otherwise the functions : `on_double_tap`, `on_triple_tap` - will trigger twice.
[Code example](https://kivymd.readthedocs.io/en/latest/behaviors/touch/index.html#usage)

### DIFF
--- a/kivymd/uix/behaviors/touch_behavior.py
+++ b/kivymd/uix/behaviors/touch_behavior.py
@@ -1,51 +1,34 @@
 """
 Behaviors/Touch
 ===============
-
 .. rubric:: Provides easy access to events.
-
 The following events are available:
-
 - on_long_touch
 - on_double_tap
 - on_triple_tap
-
 Usage
 -----
-
 .. code-block:: python
-
     from kivy.lang import Builder
-
     from kivymd.app import MDApp
     from kivymd.uix.behaviors import TouchBehavior
     from kivymd.uix.button import MDRaisedButton
-
     KV = '''
     Screen:
-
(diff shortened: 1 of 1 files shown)


kivymd/uix/behaviors/touch_behavior.py
```diff
@@ -1,51 +1,34 @@
 """
 Behaviors/Touch
 ===============
-
```
HeaTTheatR, line 4: Why did you remove indentation from the documentation?

kivymd/uix/behaviors/touch_behavior.py
```diff
@@ -1,51 +1,34 @@
...
     KV = '''
     Screen:
-
         MyButton:
             text: "PRESS ME"
             pos_hint: {"center_x": .5, "center_y": .5}
     '''
-
-
```
HeaTTheatR, line 32: The double indentation before the class declaration is the PEP8 standard.

HeaTTheatR:
@Neizvestnyj Please attach some code examples that demonstrate the old behavior and behavior according to your change ...

kivymd/uix/behaviors/touch_behavior.py
Neizvestnyj: `PyCharm` did not correct me), now I will record a video and create a new pool of request

PR #1008, opened 2021-06-13
 Changed the way to search for disks for `OSX` , `Linux`, `android` in the `show_disks` function of the `MDFileManager` class, added a preview mode for disks

Neizvestnyj:
I decided to do a disk search without using the `subprocess` module, the `mount` command is more universal as for me, the new method will work even if there are more than 2 spaces in the disk name. I also added a `preview` mode for disks, and selected icons similar to `folder.png`.


HeaTTheatR:
@Neizvestnyj I find it inappropriate to add new icons to `images_path`

Neizvestnyj:
> @Neizvestnyj I find it inappropriate to add new icons to `images_path`

Remove the preview for `show_disks`? Or the icons are not very good?

HeaTTheatR:
@Neizvestnyj Why use third-party icons by default?

Neizvestnyj:
> @Neizvestnyj Why use third-party icons by default?

Yes, I just saw that the `show_disks` function does not have a `preview` and decided to add it following the example of yours. As far as I understand, this is not really necessary? I can then remove the preview.

Neizvestnyj:
@HeaTTheatR Create a new pull request or edit this one

HeaTTheatR:
> Create a new pull request or edit this one

Create a new PR...

PR #1009, opened 2021-06-13
Changed the way to search for disks for 'OSX', 'Linux' and 'android' …

Neizvestnyj:
I decided to do a disk search without using the subprocess module, the mount command is more universal as for me, the new method will work even if there are more than 2 spaces in the disk name.

### DIFF
--- a/kivymd/uix/filemanager.py
+++ b/kivymd/uix/filemanager.py
@@ -472,31 +472,30 @@ class MDFileManager(ThemableBehavior, MDRelativeLayout):
         return sorted_files
 
     def show_disks(self):
-        if not self.disks:
-            if platform == "win":
-                self.disks = sorted(
-                    re.findall(
-                        r"[A-Z]+:.*$",
-                        os.popen("mountvol /").read(),
-                        re.MULTILINE,
-                    )
+        if platform == "win":
+            self.disks = sorted(
+                re.findall(
+                    r"[A-Z]+:.*$",
+                    os.popen("mountvol /").read(),
+                    re.MULTILINE,
                 )
-            elif platform in ["linux", "android"]:
-                self.disks = sorted(
-                    re.findall(
-                        r"on\s(/.*)\stype",
-                        os.popen("mount").read(),
-                    )
+            )
+        elif platform in ["linux", "android"]:
+            self.disks = sorted(
+                re.findall(
+                    r"on\s(/.*)\stype",
+                    os.popen("mount").read(),
                 )
-            elif platform == "macosx":
-                self.disks = sorted(
-                    re.findall(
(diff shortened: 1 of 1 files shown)


kivymd/uix/filemanager.py
```diff
@@ -653,8 +658,8 @@ def get_content(self):
                                 pass
                         else:
                             if (
-                                not self.show_hidden_files
-                                and content.startswith(".")
+                                    not self.show_hidden_files
+                                    and content.startswith(".")
```
HeaTTheatR: This is not according to the `PEP8` standard.

kivymd/uix/filemanager.py
```diff
@@ -636,7 +641,7 @@ def get_content(self):
                 if os.path.isdir(os.path.join(self.current_path, content)):
                     if self.search == "all" or self.search == "dirs":
                         if (not self.show_hidden_files) and (
-                            content.startswith(".")
+                                content.startswith(".")
```
HeaTTheatR: This is not according to the `PEP8` standard.

kivymd/uix/filemanager.py
```diff
@@ -557,8 +562,8 @@ def show(self, path):
                 )
             for name_file in self.__sort_files(files):
                 if (
-                    os.path.splitext(os.path.join(path, name_file))[1]
-                    in self.ext
+                        os.path.splitext(os.path.join(path, name_file))[1]
+                        in self.ext
```
HeaTTheatR: This is not according to the `PEP8` standard.

kivymd/uix/filemanager.py
```diff
@@ -425,9 +423,9 @@ def __init__(self, **kwargs):
         toolbar_label = self.ids.toolbar.children[1].children[0]
         toolbar_label.font_style = "Subtitle1"
         if (
-            self.selector == "any"
-            or self.selector == "multi"
-            or self.selector == "folder"
+                self.selector == "any"
+                or self.selector == "multi"
+                or self.selector == "folder"
```
HeaTTheatR: This is not according to the `PEP8` standard.

kivymd/uix/filemanager.py
Neizvestnyj: Hmm, PyCharm did it, not me)

kivymd/uix/filemanager.py
Neizvestnyj: I use autoformatting in my compiler, I will change it now

HeaTTheatR: https://github.com/kivymd/KivyMD#setup-environment

Neizvestnyj:
All fixed

PR #1037, opened 2021-07-14
Fix bug where visible attribute is checked on wrong object type

david-steffen:
When function anim_color_normal or anim_color_active is called on MDNavigationRail, the visible property is checked on MDNavigationRailItem causing attribute error

### Description of Changes
* changed item.visible to self.visible in anim_color_normal and  anim_color_active functions

### DIFF
--- a/kivymd/uix/navigationrail.py
+++ b/kivymd/uix/navigationrail.py
@@ -678,7 +678,7 @@ class MDNavigationRail(MDCard):
             t=self.color_transition,
             d=self.color_change_duration,
         ).start(item)
-        if item.visible == "Selected":
+        if self.visible == "Selected":
             Animation(
                 text_color=(0, 0, 0, 0),
                 t=self.color_transition,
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@david-steffen I don't see any error in the code for which you made the change. Perhaps you need to show the code that reproduces the error you are talking about.

david-steffen:
In kv lang it works fine but when creating a navigation rail in python the attribute error occurs. code below

```
from kivymd.uix.boxlayout import MDBoxLayout

from kivymd.app import MDApp


class Test(MDApp):
    def build(self):
        from kivymd.uix.navigationrail import MDNavigationRail, MDNavigationRailItem
        nav_rail = MDNavigationRail()
        nav_rail.visible="Unlabeled"
        nav_rail.use_action_button = True
        nav_rail_item = MDNavigationRailItem(navigation_rail=nav_rail, icon="launch", text="Open")
        nav_rail.add_widget(nav_rail_item)
        root = MDBoxLayout()
        root.add_widget(nav_rail)
        return root


Test().run()
```

PR #1049, opened 2021-07-24
Addition of 'sec_icon'

sonam-gr8:
sec_icon:'<icon_name> changes the icon situated besides the text_title

I could not find a way to change the cog icon which appeared beside the title in MDNavigationRail
therefore devised a way to use the icon

use of sec_icon: will allow the user to change the secondary icon


HeaTTheatR:
@sonam-gr8 https://github.com/kivymd/.github/blob/main/.github/pull_request_template.md

PR #1061, opened 2021-08-03
Modify Chip.py "Check" icon

SpecDaTurtle:
This modification will allow it to match the set icon color.

The "check" icon created in the "on_touch_down" function remains black regardless of what icon color is chosen for the chip.
The example MDChips below should produce a white check rather than a black one.

### Reproducing the problem

```python
MDChip:
    text: 'On Standby?'
    pos_hint: {"center_x":0.5}
    check: True
    icon_color: 1,1,1,1
    text_color: 1,1,1,1

MDChip:
    text: 'On Standby?'
    pos_hint: {"center_x":0.5}
    icon: 'account-clock'
    check: True
    icon_color: 1,1,1,1
    text_color: 1,1,1,1
```

### Screenshot of the problem

Chip problem

### Description of Changes

I included the following lines of code to the "Check" widget so that it inherits the icon color and therefore will always match it.
```python
theme_text_color = "Custom",
text_color = self.icon_color
```
### Screenshot of the solution to the problem

Solution

### DIFF
--- a/kivymd/uix/chip.py
+++ b/kivymd/uix/chip.py
@@ -296,6 +296,8 @@ class MDChip(ThemableBehavior, ButtonBehavior, BoxLayout):
                             size_hint=(None, None),
                             size=("26dp", "26dp"),
                             font_size=sp(20),
+                            theme_text_color = "Custom",
+                            text_color = self.icon_color
                         )
                     )
                 else:


SpecDaTurtle:
I wanted to split the check and normal icon up so they could have their own colours, however this does not match the material design specs as far as I can tell. For this reason I figured it would be best to link their color selection.

HeaTTheatR:
@SpecDaTurtle 
Better solution might be to add a new property to `MDChip` to make the check mark color customizable independently from either text or icon color.

HeaTTheatR:
@SpecDaTurtle 
I will close this PR as it conflicts with the main branch. You can make the appropriate changes taking into account the comment above and open a new PR.

SpecDaTurtle:
Yes, I saw someone else is busy with that exact thing so considering to leave it for another week or so to give them the chance to finish up before i fix this pr

PR #1062, opened 2021-08-08
Fix `TypeError` in `label_check_texture_size` method

Neizvestnyj:
To indicate that the argument can be of different types, you must use `Union`

### DIFF
--- a/kivymd/toast/kivytoast/kivytoast.py
+++ b/kivymd/toast/kivytoast/kivytoast.py
@@ -38,7 +38,7 @@ KivyToast
     Test().run()
 """
 
-from typing import NoReturn
+from typing import NoReturn, Union
 
 from kivy.animation import Animation
 from kivy.clock import Clock
@@ -87,7 +87,7 @@ class Toast(BaseDialog):
         self.add_widget(self.label_toast)
 
     def label_check_texture_size(
-        self, instance_label: Label, texture_size: list[int, int]
+        self, instance_label: Label, texture_size: Union[int, int]
     ) -> NoReturn:
         """
         Resizes the text if the text texture is larger than the screen size.


kivymd/toast/kivytoast/kivytoast.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, **kwargs):
         self.add_widget(self.label_toast)
 
     def label_check_texture_size(
-        self, instance_label: Label, texture_size: list[int, int]
+        self, instance_label: Label, texture_size: Union[int, int]
```
HeaTTheatR, line 90: It is not right. `texture_size` takes a list of two numbers. The `Union` type is not appropriate here.

Neizvestnyj: But `list` is incorrect, I get error

Neizvestnyj: Python 3.9.6 has a `Sequence`, but few people use this version of python. It can check arguments already directly in the function?

HeaTTheatR: No. Annotation of types only for programmers and IDEs

Neizvestnyj: But this causes an error( Maybe `List[int]`?

HeaTTheatR: I am not getting any error ...

HeaTTheatR: Maybe List(...)

Neizvestnyj: Because you have python > 3.9

Neizvestnyj: > Maybe List(...)
? I think 
```
from typing import List
List[int]
``` 
It will be correct, although there will be no limit on the number of arguments

Neizvestnyj: It's just that you use python 3.9.6 typing, and this version is not worth much to anyone...

HeaTTheatR: I see no reason to support older versions of Python. And as far as I know, Python doesn't keep track of types. You can write whatever you want in annotations.

Neizvestnyj: Tracks, in python 3.8.5 I get this error - 
```
self, instance_label: Label, texture_size: list[int, int]
TypeError: 'type' object is not subscriptable
```

HeaTTheatR:
```python
def func(var: int):
    print(var)


func("var")
```

This code will not throw any errors because Python doesn't keep track of the types of the arguments!

Neizvestnyj:
> ```python
> def func(var: int):
>     print(var)
> 
> 
> func("var")
> ```
> 
> This code will not throw any errors because Python doesn't keep track of the types of the arguments!

Maybe now (in 3.9), I get error

HeaTTheatR:
@Neizvestnyj 


https://user-images.githubusercontent.com/16930280/128646126-673df498-8855-4864-8241-26bb7532af46.mov

Neizvestnyj:
> @Neizvestnyj
> 
>  2021-08-09.00.22.07.mov

This is due to the fact that you specify as a type a `list` in which `int`s, and the old python, tries to pass them, because of this error

HeaTTheatR:
@Neizvestnyj Perhaps need to use the List class from the typing module

Neizvestnyj:
I checked, there is no such error in python 3.9.6. Just specify in the dependencies that you need python version 3.9. (`requirements = kivy==2.0.0, kivymd==0.104.2, sdl2_ttf == 2.0.15, pillow, python3==3.9.6`)

HeaTTheatR:
@Neizvestnyj It is necessary to indicate this in README that we do not support Python <3.9

PR #1063, opened 2021-08-08
The supported python version has been changed to 3.9

Neizvestnyj:


HeaTTheatR:
@Neizvestnyj 
Снимок экрана 2021-08-09 в 00 38 35

Neizvestnyj:
> @Neizvestnyj
> Снимок экрана 2021-08-09 в 00 38 35

But the user does not start on 3.6-3.8, I checked, you can also create a honda virtual environment and check

HeaTTheatR:
@Neizvestnyj I say that these icons need to be removed except for 3.9

Neizvestnyj:
> @Neizvestnyj I say that these icons need to be removed except for 3.9

Ah)

Neizvestnyj:
> @Neizvestnyj I say that these icons need to be removed except for 3.9

As I know, to update image, we need to set in `setup.py`:
```
classifiers=[
        'Programming Language :: Python',
        'Programming Language :: Python :: 3.9',
    ],
 ```

Neizvestnyj:
But it seems a little strange to me, to be deprived of support for old versions of python due to typing in one class ... We could just use `List`, but then the programmer would not immediately understand how many arguments this list takes, but as for me, this is not so important

Issue #1116, opened 2021-11-01
MDRaisedButton critical clock


Amyantis:
### Description of the Bug

Reopening https://github.com/kivymd/KivyMD/issues/1011

I think the warning `[CRITICAL] [Clock ] Warning, too much iteration done before the next frame. Check your code, or increase the Clock.max_iteration attribute` shows if:

The button is added to a screen that is not currently displayed.
The button is a child of a BoxLayout with padding>0.

Here is a reproducible example:

### Code and Logs

```python
from kivy.lang import Builder
from kivymd.app import MDApp
from kivymd.uix.screen import MDScreen


class ScreenA(MDScreen):
    pass


class ScreenB(MDScreen):
    pass


KV = """
<ScreenA>:
    BoxLayout:
        orientation: 'vertical'
        padding: 10
        MDLabel:
            text: "Hello world A!"
        MDRaisedButton:
            text: "Go world B"
            size_hint_x: 1.0
            on_press: app.root.ids.screen_manager.current = "ScreenB"
            
<ScreenB>:
    BoxLayout:
        orientation: 'vertical'
        padding: 10
        MDLabel:
            text: "Hello world B!"
        MDRaisedButton:
            text: "Go world A"
            size_hint_x: 1.0
            on_press: app.root.ids.screen_manager.current = "ScreenA"
            
Screen:
    ScreenManager:
        id: screen_manager
        ScreenA:
            name: "ScreenA"
        ScreenB:
            name: "ScreenB"
"""


class MyMDApp(MDApp):
    def build(self):
        return Builder.load_string(KV)


if __name__ == "__main__":
    MyMDApp().run()

```

### Versions

* OS: Ubuntu 20.04.3
* Python: 3.8.10
* Kivy: 2.0.0
* KivyMD: 0.104.2

Amyantis:
Here is the dirty workaround I use at the moment:

```python
class CustomMDScreen(MDScreen):
    are_button_already_resized = BooleanProperty(False)

    def on_pre_enter(self, *args):
        super().on_pre_enter(*args)
        if self.are_button_already_resized:
            return
        for widget in self.walk():
            if isinstance(widget, MDRaisedButton):
                widget.size_hint_x = 1.0
        self.are_button_already_resized = True
```

gbm001:
I am suffering from the same problem... although I just started working on a possible refactoring of the button code and I don't know if the problem persists in that version.

gbm001:
[...]

PR #1141, opened 2021-12-14
fix(buttons): Fix infinite loop for buttons with minimum width

gbm001:

As described in #1116 , creating a button that:
- has a minimum width set (any rectangular button),
- has a non-zero size_hint_x,
- is created in a BoxLayout with a non-zero padding, and
- is on a Screen that is not yet displayed,
causes an immediate infinite loop.

The on_width of BaseButton sets the width to 88. However, AnchorLayout then sets this width to a different value; this may also trigger updates in BoxLayout and/or the width binding in ButtonContentsText (Kivy language). The set width (for a button on a screen that has not yet been shown) is less than 88, triggering on_width again hence the infinite loop.

### Reproducing the problem

Adapted from #1116

```python
from kivy.lang import Builder
from kivymd.app import MDApp
from kivymd.uix.screen import MDScreen


class ScreenA(MDScreen):
    pass


class ScreenB(MDScreen):
    pass


KV = """
<ScreenA>:

<ScreenB>:
    BoxLayout:
        padding: 10
        MDFlatButton:
            size_hint_x: 1.0

Screen:
    ScreenManager:
        ScreenA:
            name: "ScreenA"
        ScreenB:
            name: "ScreenB"
"""


class MyMDApp(MDApp):
    def build(self):


if __name__ == "__main__":
    MyMDApp().run()
```

image

### Description of Changes

In BaseButton __init__(), check if _min_width is set. If it is, then set size_hint_min_x to the larger of _min_width and size_hint_min_x (if set). This will normally just set size_hint_min_x to _min_width, but in the event a user sets size_hint_min_x to a larger value then this will be preserved.

The layout code in AnchorLayout now keeps the size of the button above _min_width, preventing the infinite loop.

### DIFF
--- a/kivymd/uix/button/button.kv
+++ b/kivymd/uix/button/button.kv
@@ -32,7 +32,7 @@
 
 <ButtonContentsText>
     lbl_txt: lbl_txt
-    width: lbl_txt.texture_size[0] + dp(24)
+    width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
     height: dp(20) + lbl_txt.texture_size[1]
 
     MDLabel:
@@ -74,7 +74,7 @@
 <ButtonContentsIconText>
     lbl_txt: lbl_txt
     lbl_ic: lbl_ic
-    width: lbl_txt.width + lbl_ic.width * 2 + box.spacing
+    width: max(root._min_width, lbl_txt.width + lbl_ic.width * 2 + box.spacing)
     height: lbl_ic.texture_size[1] + dp(12)
 
     MDBoxLayout:
--- a/kivymd/uix/button/button.kv
+++ b/kivymd/uix/button/button.kv
@@ -33,6 +33,7 @@
 <ButtonContentsText>
     lbl_txt: lbl_txt
     width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
+    size_hint_min_x: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
     height: dp(20) + lbl_txt.texture_size[1]
 
     MDLabel:
@@ -75,6 +76,7 @@
     lbl_txt: lbl_txt
     lbl_ic: lbl_ic
     width: max(root._min_width, lbl_txt.width + lbl_ic.width * 2 + box.spacing)
+    size_hint_min_x: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
     height: lbl_ic.texture_size[1] + dp(12)
 
     MDBoxLayout:
--- a/kivymd/uix/button/button.py
+++ b/kivymd/uix/button/button.py
@@ -699,13 +699,6 @@ class BaseButton(
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        # To prevent infinite loop for partially-initialized button width
-        # between AnchorLayout and on_width, set size_hint_min_x to match
-        # (at least) _min_width.
-        if self._min_width:
-            self.size_hint_min_x = max(
-                self._min_width, self.size_hint_min_x or 0
-            )
         self.theme_cls.bind(
             primary_palette=self.set_button_colors,
             theme_style=self.set_button_colors,
@@ -787,14 +780,6 @@ class BaseButton(
         else:
             default_text_color = self.theme_cls.text_color
         self._text_color = self.text_color or default_text_color
-
-    def on_width(self, instance_button, width: float) -> NoReturn:
-        """
-        If the button style has a minimum width set, enforce it here.
-        """
-
-        if self._min_width and (self.width < self._min_width):
-            self.width = self._min_width
 
     # Touch events that cause transparent buttons to fade to background
     def on_touch_down(self, touch):
@@ -931,7 +916,7 @@ class ButtonContentsIcon(ButtonIconMixin):
     Contents for a round BaseButton consisting of an MDIcon.
     """
 
-    _min_width = None
+    _min_width = 0
 
     def on_text_color(self, instance_button, color: list) -> NoReturn:
         """
@@ -1167,7 +1152,7 @@ class MDIconButton(ButtonContentsIcon, BaseButton):
     and defaults to `None` (set by button class).
     """
 
-    _min_width = None
+    _min_width = 0
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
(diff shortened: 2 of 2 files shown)


HeaTTheatR:
@gbm001 These changes do not fix the problem.

gbm001:
> @gbm001 These changes do not fix the problem.

Could you define the problem that this doesn't fix?

Using the current master branch, running the test code provided in this PR results in the infinite looping shown. My testing shows that on_width is being called repeatedly to set a width of 80, which then gets set to 88 by on_width, and the cycle repeats.

image

Using the extra lines in this PR, no infinite loop occurs.

image

Arguably, using on_width to force a minimum width is the problem, and could/should have been implemented in a better way (e.g. the Kivy language binding that sets button width could be set to max(self._min_width, [appropriate expression]), but that's a bigger change.

gbm001:
Using on_width is fragile if anything else tries to set a width (and I've just stumbled over that myself a few times when trying to force widths on buttons in my own app), so I can imagine there are still issues in different cases, but this appears to work fine for both the buttons in my App (which had the same issue as the original issue submitter), and the test case provided by them (cut down by me)?

HeaTTheatR:
@gbm001 Doesn't fix this problem - https://github.com/kivymd/KivyMD/issues/1116

gbm001:
Apologies, you are correct.
Turns out there is some difference between my cut-down test and the original test, although I'm not sure what...
Probably just getting rid of on_width is the better solution or we'll keep having issues with this.

gbm001:
Let's try that again...

The new solution is that the on_width method is deleted, removing the infinite loop.
To enforce the minimum width, the Kivy language bindings that set the button width note the minimum width instead, e.g.:
```python
width: lbl_txt.texture_size[0] + dp(24)
```
becomes
```python
width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
```

This doesn't stop users setting their buttons to something less than the specified min_width, but it does stop the button becoming more than the specified min_width due to short contents (e.g. a short label).

That my old 'fix' worked for my own app and my cut-down version of the original test, but not the original test, means the fix (and the problem) was fragile. This solution should be less fragile (hopefully).

HeaTTheatR:
> Let's try that again...
> 
> The new solution is that the on_width method is deleted, removing the infinite loop. To enforce the minimum width, the Kivy language bindings that set the button width note the minimum width instead, e.g.:
> 
> ```python
> width: lbl_txt.texture_size[0] + dp(24)
> ```
> 
> becomes
> 
> ```python
> width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
> ```
> 
> This doesn't stop users setting their buttons to something less than the specified min_width, but it does stop the button becoming more than the specified min_width due to short contents (e.g. a short label).
> 
> That my old 'fix' worked for my own app and my cut-down version of the original test, but not the original test, means the fix (and the problem) was fragile. This solution should be less fragile (hopefully).

The button must be no less than 88 pixels.

HeaTTheatR:
@gbm001 

Снимок экрана 2021-12-14 в 20 39 13

This behavior should not suit us.

gbm001:
> > Let's try that again...
> > The new solution is that the on_width method is deleted, removing the infinite loop. To enforce the minimum width, the Kivy language bindings that set the button width note the minimum width instead, e.g.:
> > ```python
> > width: lbl_txt.texture_size[0] + dp(24)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > becomes
> > ```python
> > width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This doesn't stop users setting their buttons to something less than the specified min_width, but it does stop the button becoming more than the specified min_width due to short contents (e.g. a short label).
> > That my old 'fix' worked for my own app and my cut-down version of the original test, but not the original test, means the fix (and the problem) was fragile. This solution should be less fragile (hopefully).
> 
> The button must be no less than 88 pixels.

Indeed, and that is why for all the rectangular buttons (namely anything other than MDIconButton and MDFloatingActionButton) _min_width is defined (in BaseButton) as 88.

HeaTTheatR:
> Indeed, and that is why for all the rectangular buttons (namely anything other than MDIconButton and MDFloatingActionButton) _min_width is defined (in BaseButton) as 88.

This does not work:

```python
from kivy.lang import Builder

from kivymd.app import MDApp


KV = """
MDScreen:

    MDRaisedButton:
        text: "MDRaisedButton"
        text_color: 0, 1, 1, 1
        size_hint_x: .05
"""


class MyMDApp(MDApp):
    def build(self):


MyMDApp().run()
```

Снимок экрана 2021-12-14 в 20 39 13

gbm001:
It's like whacking moles...

I had checked the Kitchen Sink demo, your test buttons script, and my own test script which had some short button labels (which were definitely being enlarged to 88dp)... I presume it's the size_hint parameter that breaks it. Will attempt to re-fix...

gbm001:
OK, so setting size_hint_min_x together with width solves the problem in that case (button size set by size_hint).
image

I'll just do a bit more testing to check the same applies when creating the button from Python.

gbm001:
Seems to work from Python as well.

image

image


```python
from kivy.lang import Builder

from kivymd.app import MDApp
from kivymd.uix.button import MDRaisedButton


KV = """
MDScreen:

    BoxLayout:
        size_hint: 1, 1
        orientation: "horizontal"

        RelativeLayout:
            size_hint: 0.5, 1

            MDRaisedButton:
                text: "Very Long Text Label That Goes On"
                text_color: 0, 1, 1, 1
                size_hint_x: .05

        RelativeLayout:
            id: python_box
            size_hint: 0.5, 1
"""


class MyMDApp(MDApp):
    def build(self):

    def on_start(self):
        self.root.ids.python_box.add_widget(
            MDRaisedButton(
                text='Very Long Text Label That Goes On',
                text_color=[0, 1, 1, 1],
                size_hint_x=0.05,
                pos_hint={'center_x': 0.5, 'center_y': 0.5}
            )
        )


MyMDApp().run()
```

gbm001:
thanks :)

HeaTTheatR:
@gbm001 Great fix!

PR #1143, opened 2021-12-16
fix(buttons): Changing the elevation of a disabled button displays a shadow

gbm001:
Setting the elevation of an MDFloatingActionButton or MDRaisedButton while
the button is disabled no longer causes the shadow effect to be displayed
when it should not be.

The property 'elevation' can be used to set the elevation of an elevation-supporting KivyMD button class (currently MDRaisedButton and MDFloatingActionButton). However, if the elevation is changed while the button is disabled, elevation effects (namely the background shadow) are disabled. Disabled buttons should not display any elevation behaviour. Changing the disabled status (e.g. toggling between disabled and un-disabled) fixes the problem, as the method on_disabled in CommonElevationBehaviour properly clears all animations etc.

### Reproducing the problem

```python
from kivy.lang import Builder
from kivymd.app import MDApp
from kivymd.uix.label import MDLabel
from kivymd.uix.button import MDFlatButton, MDFloatingActionButton, MDRaisedButton


KV = """
MDScreen:

    ScrollView:

        MDBoxLayout:
            id: box
            orientation: "vertical"
            spacing: dp(10)
            adaptive_size: True

            MDLabel:
                text: "Elevation 6"
                adaptive_size: True

            MDFloatingActionButton:
                id: kv_1
                icon: "android"
                elevation: 6
                disabled: True

            MDRaisedButton:
                id: kv_2
                text: "PRESS"
                elevation: 6
                disabled: True

            MDFlatButton:
                text: "TOGGLE DISABLE"
                on_release:
                    kv_1.disabled = not kv_1.disabled
                    kv_2.disabled = not kv_2.disabled

            MDFlatButton:
                text: "INCREASE ELEVATION"
                on_release:
                    kv_1.elevation = kv_1.elevation + 4
                    kv_2.elevation = kv_2.elevation + 4

            MDFlatButton:
                text: "DECREASE ELEVATION"
                on_release:
                    kv_1.elevation = max(2, kv_1.elevation - 4)
                    kv_2.elevation = max(2, kv_2.elevation - 4)
"""

class Test(MDApp):
    def build(self):


Test().run()
```

Before increasing elevation
image

After increasing elevation
image

### Description of Changes

Changes are made in ButtonElevationBehavior.
1. on_disabled is not called at the end of __init__, as it is now called at the end of on_elevation.
2. The line `self._anim_raised = Animation(_elevation=self._elevation_raised, d=0.15)` is removed as it is redundant; the same call is made in on__elevation_raised which will always be triggered immediately before this line anyway.
3. on_disabled is always called at the end of on_elevation.

After clicking 'INCREASE ELEVATION' or 'DECREASE ELEVATION', no visual change is observed in disabled widgets.
image

### DIFF
--- a/kivymd/uix/button/button.py
+++ b/kivymd/uix/button/button.py
@@ -946,12 +946,11 @@ class ButtonElevationBehaviour(CommonElevationBehavior):
         super().__init__(**kwargs)
         self.bind(_radius=self.setter("radius"))
         self.on_elevation(self, self.elevation)
-        self.on_disabled(self, self.disabled)
 
     def on_elevation(self, instance_button, elevation_value: int) -> NoReturn:
         super().on_elevation(instance_button, elevation_value)
         self._elevation_raised = self.elevation + 6
-        self._anim_raised = Animation(_elevation=self._elevation_raised, d=0.15)
+        self.on_disabled(self, self.disabled)
 
     def on__elevation_raised(
         self, instance_button, elevation_value: int


HeaTTheatR:
@gbm001 https://github.com/kivymd/KivyMD/discussions/1145?converting=1

PR #1146, opened 2021-12-20
fix: Fix many issues with MD2 button styling

gbm001:

There are a number of issues with styling of buttons in accordance with the Material Design specs (version 2). There are also some fixes for Floating Action Buttons for version 3 of the MD specs.

### Changes

Unless otherwise stated, all references to the Material Design (MD) specifications refer to version 2.

#### Enhancing BaseButton for MD3
In readiness for developing a single MDButton class to handle all (rounded) rectangular MD buttons in both version 2 and 3 (with MDIconButton and MDFloatingActionButton remaining separate), the 'icon' related properties are moved into BaseButton. Some mix-in classes are rearranged/renamed and altered to support the existing series of MD button classes. A new attribute 'RoundedButton' is used to state that a button is to have fully-rounded corners i.e. the radius is half the height. This is used for the existing 'Round' button classes, the circular classes (MDIconButton and MDFloatingActionButton) and can be used for MD3 buttons (when implemented).

#### Width
The MD spec requires that the minimum width of a button (with an outline) is 64dp. In KivyMD, this was set to 88px. Furthermore, the 'padding' attribute was not really used.
The default padding values and the 'width' declarations in Kivy language have been changed to match the MD spec of 16dp for outlined/filled buttons (only 12dp on the left if there is a left icon) or 8dp each side for text buttons.
For buttons with an icon and text, the spacing is reduced from 8dp to 4dp as required by the MD spec.

#### Height
The MD spec requires the height of a button to be 36dp. We may have taller buttons, so we set the minimum height to 36dp and have reduced the vertical padding to 8dp (top and bottom) which, for the default font size, creates a 36dp button. Larger font sizes will create larger buttons than the MD spec allows to accommodate the text.

#### Radius
The rounded radius of buttons is now, by default, 4dp as specified in the MD spec. For 'rounded' buttons (ones where rounded_button is set True) the rounded radius is now set to half the height, and this is also used for MDIconButton and MDFloatingActionButton to create circular buttons.

#### Text centering
Both text and icons (where present) are vertically centered within buttons (for the rectangular buttons) as the font size of each can be controlled separately. Furthermore, halign is set to 'center' and valign to 'middle' for all text labels so that text is correctly positioned for multi-line labels (although these are not permitted under the MD spec).

#### Icon font size
The icon font size can be set explicitly using 'icon_size', but when not set it is now set to 18/14 * text label font size (for rectangular buttons). This gives the MD spec specified 18dp icon when the default 14sp text is used, and keeps the icon and text in better size agreement than the previous 'root font size + sp(14)'. Using the label font size also deals with the button have font_style but not font_size specified.

#### Floating Action Button sizes
M3 floating action button rounded corner sizes were specified in pixels, rather than dp. They have now been set to the correct sizings.

### DIFF
--- a/kivymd/uix/button/button.kv
+++ b/kivymd/uix/button/button.kv
@@ -32,15 +32,26 @@
 
 <ButtonContentsText>
     lbl_txt: lbl_txt
-    width: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
-    size_hint_min_x: max(root._min_width, lbl_txt.texture_size[0] + dp(24))
-    height: dp(20) + lbl_txt.texture_size[1]
+    width:
+        max(root._min_width, \
+        root.padding[0] + lbl_txt.texture_size[0] + root.padding[2])
+    size_hint_min_x:
+        max(root._min_width, \
+        root.padding[0] + lbl_txt.texture_size[0] + root.padding[2])
+    height:
+        max(root._min_height, \
+        root.padding[1] + lbl_txt.texture_size[1] + root.padding[3])
+    size_hint_min_y:
+        max(root._min_height, \
+        root.padding[1] + lbl_txt.texture_size[1] + root.padding[3])
 
     MDLabel:
         id: lbl_txt
         text: root.text
         font_size: root.font_size
         font_style: root.font_style
+        halign: 'center'
+        valign: 'middle'
         adaptive_size: True
         -text_size: None, None
         theme_text_color: root._theme_text_color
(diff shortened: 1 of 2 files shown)


kivymd/uix/button/button.kv
```diff
@@ -32,15 +32,26 @@
...
+        max(root._min_height, \
+        root.padding[1] + lbl_txt.texture_size[1] + root.padding[3])
 
     MDLabel:
         id: lbl_txt
         text: root.text
         font_size: root.font_size
         font_style: root.font_style
+        halign: 'center'
+        valign: 'middle'
```
HeaTTheatR, line 54: This will not allow changing the position of the text.

gbm001: You can't do that anyway? (well not easily)
This only affects multi-line text. If you want to move the whole text label around within the button, the existing halign and valign on the _button_, not the label, would be the things to do that with (although I don't know if they actually work). It would, I think, also not be in accordance with the MD spec.

gbm001:
I'm going to put up screenshots of the changes in future comments.

gbm001:
Width changes and radius changes.

Before:
image
There is no rounding of corners. The smallest button width allowed is 88px.

After
image
Corners are rounded to the MD spec of 4dp. The smallest button width allowed is 64dp.

This is on a PC; on a phone 64dp could be bigger than 88px.

(ignore the spacing between the buttons which is just how much I shrunk the window)

gbm001:
Height changes and icon size changes
Normally the changes to the way height is calculated shouldn't make any difference - default buttons remain at 36dp height.
Icon sizes are now more in accordance with text sizes, when not specified explicitly.

Before:
image
Icon sizes are not in accordance with the text size, for a range of font sizes.
In these cases, font size on the _button_ remains at the default 14sp (as it is not explicitly set) even when different font_styles are used (setting different font_sizes on the text _label_) and thus the icon size is 14 + 14sp.

After
image
Icon sizes are in accordance with the text size, with the MD spec default of 18sp for the default font size of 14sp. Icon font sizes are 18/14 * font_size on the _label_ (as icon_size on the _button_ is not explicitly set).
Button heights are the same as before.

Also notice that the corners are correctly fully-rounded for the larger buttons; previously the rounding radius was set to 18dp regardless of button height.

gbm001:
Text label alignment

Before
image

After
image

Although the label as a whole is centred within the button, in the first case the text within the label is not (as halign is, by default, effectively 'left' for a Label). In the second case, the multi-line text is correctly centred (albeit such labels are not permitted under the MD spec).

kivymd/uix/button/button.kv
HeaTTheatR: > You can't do that anyway? (well not easily) This only affects multi-line text. If you want to move the whole text label around within the button, the existing halign and valign on the _button_, not the label, would be the things to do that with (although I don't know if they actually work). It would, I think, also not be in accordance with the MD spec.

It worked. But you removed this feature in your previous changes.

gbm001: See the last set of screenshots which show how halign on the _label_, not the _button_, affect the positioning of multi-line text.

gbm001: > It worked. But you removed this feature in your previous changes.

halign and valign on the _button_ still work:

image

https://pastebin.com/ULWSaaSV

kivymd/uix/button/button.py
```diff
@@ -611,24 +651,23 @@ class BaseButton(
     and defaults to `14sp`.
     """
 
-    user_font_size = NumericProperty(0)
```
HeaTTheatR, line 614: This will entail a bunch of mistakes. And for many users and in the demo application! You cannot just remove this attribute. You should marked it as deprecated. And make a `user_font_size` method that passes its value to the `icon_size` attribute.

gbm001: It's not gone, just moved - it's still an attribute of BaseButton (nothing has really changed). The order of attributes changed slightly when the icon properties were merged into BaseButton. In fact I haven't done anything with user_font_size which was only used with MDIconButton and MDFloatingActionButton.

But yes, it should be deprecated and replaced with icon_size.

PR #1158, opened 2022-01-04
minor typo fix for screen module

cesar-pz:

Fixed module-related typos for MDScreen usage so readers can copy and paste without worry.


HeaTTheatR:
There are no such imports in the kivymd.theming.py module.

PR #1159, opened 2022-01-07
Fixed scrolling issue of header and made scrolling easier

Thanish-Kumar:

1. Scrolling was happening only when content is dragged and not when the scroll bar is dragged.
2. Header of the table was not scrolling

### Reproducing the problem

```python
# Minimal code example that reproduces the problem:
```

Attach screenshots that demonstrate the problem.

### Description of Changes
1. Changed option to scroll when both content or scroll bar is dragged.
2. Added a function to update scroll value of header along with table rows

Describe what changes you create to solve the problem.

Attach screenshots that demonstrate that your changes solved the problem.

### DIFF
--- a/kivymd/uix/datatables/datatables.kv
+++ b/kivymd/uix/datatables/datatables.kv
@@ -100,7 +100,7 @@
 
 <TableHeader>
     bar_width: 0
-    do_scroll: False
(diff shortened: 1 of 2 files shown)


HeaTTheatR:

PR #1168, opened 2022-01-13
Added new class `MDRecycleGridLayout` to simplifies working with some widget properties

Neizvestnyj:
```py
from kivymd.app import MDApp

KV = """
Screen:
    RecycleView:
        id: rv
        viewclass: 'OneLineAvatarIconListItem'
        MDRecycleGridLayout:
            adaptive_height: True
            default_size_hint: 1, None
            default_size: None, dp(48)
            cols: 1
            md_bg_color: app.theme_cls.primary_color
"""


class TestApp(MDApp):
    def build(self):

    def on_start(self):
        widgets = []

        for i in range(20):
            widgets.append({'text': str(i)})

        self.root.ids.rv.data = widgets


TestApp().run()

```


HeaTTheatR:

PR #1183, opened 2022-02-12
Added `reset_focus_on_text_empty` to `MDTextField` class

Neizvestnyj:
A new argument `reset_focus_on_text_empty` has been added to give the user more flexibility to control the text field. In most cases, if the user erases the text, then he does not want the text field to be defocused (for example, he enters a password and made a mistake, therefore he presses the backspace, and the field is defocused because of this, which he does not want)

```py
from kivy.lang import Builder

from kivymd.app import MDApp


KV = '''
BoxLayout:
    padding: "10dp"

    MDTextField:
        id: text_field_error
        reset_focus_on_text_empty: False
        hint_text: f'Reset focus on text empty: {self.reset_focus_on_text_empty}'
        pos_hint: {"center_y": .5}
'''


class Test(MDApp):
    def build(self):

Test().run()

https://user-images.githubusercontent.com/40869738/153708341-f4d2461a-cb1e-4a01-873e-8d00f44d574e.mp4


```

### DIFF
--- a/kivymd/uix/textfield/textfield.py
+++ b/kivymd/uix/textfield/textfield.py
@@ -926,6 +926,14 @@ class MDTextField(ThemableBehavior, TextInput):
     and defaults to `'Roboto'`.
     """
 
+    reset_focus_on_text_empty = BooleanProperty(False)
+    """
+    Is the text field unfocused if there are no characters in it.
+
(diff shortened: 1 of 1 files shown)


HeaTTheatR:

PR #1233, opened 2022-04-27
fixed bug, MDDataTable and MDTooltip

feiticeiro-tec:

Bug when selecting a row in the table while it is created
the bug is rare but it exists after several attempts at one time it happens

hover over the table while it is being created

### Reproducing the problem
```python
self.ids.table.add_widget(
MDDataTable(
check=True,
column_data=[
        ("X", dp(30)),
        ("X", dp(20)),
        ("X", dp(20))
    ],
row_data=[['1','1','1'],['1','1','1'],['1','1','1'],['1','1','1'],['1','1','1']])
```
bug-1
bug-2

### Description of Changes

try and except in datatable and tooltip

```python
kivymd/uix/datatables/datatables.py
try:
    self.ids.row_controller.select_next(self)
except IndexError:
     ...
```
```python
kivymd/uix/tooltip/tooltip.py
try:
    Window.add_widget(self._tooltip)
    pos = self.to_window(self.center_x, self.center_y)
   [....]
except WidgetException:
    ...
```

### DIFF
--- a/kivymd/uix/datatables/datatables.py
+++ b/kivymd/uix/datatables/datatables.py
@@ -500,8 +500,10 @@ class TableData(RecycleView):
 
     def set_default_first_row(self, interval: Union[int, float]) -> None:
         """Set default first row as selected."""
-
-        self.ids.row_controller.select_next(self)
+        try:
+            self.ids.row_controller.select_next(self)
+        except IndexError:
+            ...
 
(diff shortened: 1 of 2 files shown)


HeaTTheatR:
@feiticeiro-tec Using try/except blocks is not a solution to the problem. We can't use it.

Issue #1220, opened 2022-04-04
Crash upon hovering button tooltip, moving cursor outside of window, then immediately back onto button


mcdcam:
### Description of the Bug

When moving your cursor over ~~an MDTopAppBar~~ a button with a tooltip, and then out of the window, then back onto the button, there is a chance the app will crash, as the tooltip is attempted to be added to the window when it is already there, resulting in an error such as `kivy.uix.widget.WidgetException: Cannot add <kivymd.uix.tooltip.tooltip.MDTooltipViewClass object...> to window, it already has a parent <kivy.core.window.window_sdl2.WindowSDL object...>`. In normal use, this is quite rare, I probably experience it on average once per 15-20 minutes, however it can be intentionally triggered in under a few seconds.

~~I don't know why this issue is specific to MD(Top/Bottom)AppBars, but~~ the crash is happening in `tooltip.py`, on [line 204](https://github.com/kivymd/KivyMD/blob/13400aadd4a32fb58a930133d37265049d8c8578/kivymd/uix/tooltip/tooltip.py#L204), where the tooltip is added to the window (perhaps a check for whether it already has a parent is needed?).

#### EDIT:
After reducing the size of the other button, I was able to reproduce the error outside of an MDTopAppBar. It seems the bug affects all tooltips.

### Code and Logs
Minimal demo:

```python
from kivymd.app import MDApp
from kivy.lang import Builder

kv = """
<TooltipMDIconButton@MDIconButton+MDTooltip>

MDBoxLayout:
    orientation: "vertical"

    MDTopAppBar:
        # THE TOOLTIPS CAUSING THE ISSUE ARE DEFINED HERE
        title: "Title"
        left_action_items:[["arrow-left", app.callback, "Back"]]
        right_action_items: [["home", app.callback, "Home"], ["account", app.callback, "Account"]]

    MDFloatLayout:
        TooltipMDIconButton:
            icon: "language-python"
            icon_size: dp(24)
            tooltip_text: self.icon
            pos_hint: {"center_x": .06, "center_y": .08}
"""

class MainApp(MDApp):
    def callback(self, button):
        print('Callback')

    def build(self):
        self.root = Builder.load_string(kv)


if __name__ == '__main__':
    MainApp().run()
```

#### Steps to Reproduce:
- Run the above code
- Rapidly jiggle cursor over one of the ~~TopAppBar~~ buttons, moving it over the button, outside of the bounds of the window, then back over the button etc. (see video for demo)
[...]

PR #1234, opened 2022-04-28
Properly fix https://github.com/kivymd/KivyMD/issues/1220

mcdcam:

As described in https://github.com/kivymd/KivyMD/issues/1220

@HeaTTheatR made a fix, but it did not resolve the problem, as I outlined [here](https://github.com/kivymd/KivyMD/issues/1220#issuecomment-1088140635):
> `if not self._tooltip and not self._tooltip.parent:` does not prevent the crash, as it will only return if there is no tooltip *and* the tooltip does not have a parent (meaning if there is no tooltip, you will get another crash, as None has no parent), and if there is a tooltip but it already has a parent, both conditions will be false.

As described in https://github.com/kivymd/KivyMD/issues/1220

### Reproducing the problem

As described in https://github.com/kivymd/KivyMD/issues/1220

As described in https://github.com/kivymd/KivyMD/issues/1220

### Description of Changes

Changing `if not self._tooltip and not self._tooltip.parent:` to `if not self._tooltip or self._tooltip.parent:` makes the function return if there is *either* no tooltip or the tooltip already has a parent. This resolves both possible crashes in the current code, by ensuring that the function exits if it would crash by continuing.

Same code as used in https://github.com/kivymd/KivyMD/issues/1220

### DIFF
--- a/kivymd/uix/tooltip/tooltip.py
+++ b/kivymd/uix/tooltip/tooltip.py
@@ -199,7 +199,7 @@ class MDTooltip(ThemableBehavior, HoverBehavior, TouchBehavior):
         return x, y
 
     def display_tooltip(self, interval: Union[int, float]) -> None:
-        if not self._tooltip and not self._tooltip.parent:
+        if not self._tooltip or self._tooltip.parent:
             return
 
         Window.add_widget(self._tooltip)


feiticeiro-tec:
i also ended up solving this error so i used try except, it appeared after i solved another table row selection bug.
https://github.com/kivymd/KivyMD/pull/1233/files

PR #1285, opened 2022-07-16
Fix tooltip on android

Neizvestnyj:
```py
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDBoxLayout:
    orientation: "vertical"

    MDTopAppBar:
        title: "MDTopAppBar"
        right_action_items: [["dots-vertical", lambda x: print(), 'Text']]

    MDLabel:
        text: "Content"
        halign: "center"
'''


class Test(MDApp):
    def build(self):


Test().run()
```
https://user-images.githubusercontent.com/40869738/179351148-2a17eeb1-d785-4454-b5bf-3f05c10c53f4.mp4

### DIFF
--- a/kivymd/uix/tooltip/tooltip.py
+++ b/kivymd/uix/tooltip/tooltip.py
@@ -279,7 +279,7 @@ class MDTooltip(ThemableBehavior, HoverBehavior, TouchBehavior):
 
     def on_long_touch(self, touch, *args) -> None:
         if DEVICE_TYPE != "desktop":
-            self.on_enter(True)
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@Neizvestnyj I don't quite understand what problem this PR solves?

Neizvestnyj:
> @Neizvestnyj I don't quite understand what problem this PR solves?

Tooltip does not work on android, when you press the button it should appear. Here the cause `if not args and DEVICE_TYPE == "desktop":`

PR #1301, opened 2022-08-03
Fix linting errors

kulothunganug:


kivymd/tools/patterns/create_project.py
```diff
@@ -589,16 +589,16 @@ def generate_application_screens(self) -> None:
 .PHONY: po mo
 
 po:
-	xgettext -Lpython --output=messages.pot --from-code=utf-8 \\
```
HeaTTheatR, line 592: Why do you use 8 spaces instead of 4?

kulothunganug: I used 4 spaces, the line you pointed is a tab (old) and the line below is using 4-spaces (new)

Issue #1306, opened 2022-08-09
Create console scripts


JulianOrteil:
### Description of the Feature

Commands like ``python -m kivymd.tools.patterns.create_project`` are unwieldy and can be improved via console scripts. For example, the aforementioned command can be shorted to ``create_project`` via a console script.

HeaTTheatR:
@JulianOrteil 
` python kivymd/tools/release/make_release.py release 1.0.1 1.1.0.dev0 ` - https://kivymd.readthedocs.io/en/1.0.0/api/kivymd/tools/release/make_release/

HeaTTheatR:
@JulianOrteil 
`python -m kivymd.tools.patterns.add_view` - https://kivymd.readthedocs.io/en/1.0.0/api/kivymd/tools/patterns/add_view/

JulianOrteil:
@HeaTTheatR Both modules are implemented as requested. PR is ready to merge unless you have others you'd like done.

PR #1307, opened 2022-08-09
Console scripts

JulianOrteil:

#1306

### Description of Changes

Modified ``setup.py``

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -128,9 +128,9 @@ if __name__ == "__main__":
                 "tests = kivymd.tools.packaging.pyinstaller:get_pyinstaller_tests",
             ],
             "console_scripts": [
-                "add_view = kivymd.tools.patterns.add_view:main",
-                "create_project = kivymd.tools.patterns.create_project:main",
-                "make_release = kivymd.tools.release.make_release:main",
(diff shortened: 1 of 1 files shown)


kulothunganug:
Maybe `kivymd.{command}` will be more specific to kivymd lib

HeaTTheatR:
@JulianOrteil Will there be name conflicts with other scripts in this way?

JulianOrteil:
@HeaTTheatR It's certainly possible if the user has other programs or scripts on ``PATH`` with these names. I'm not partial to any naming scheme, so if you prefer something like @Kulothungan16's suggestion, that isn't an issue (albeit I've never used dot notation for console scripts so I can't comment on if that specific scheme will work on every platform).

HeaTTheatR:
@JulianOrteil Let's try to do as @Kulothungan16's suggested. The functionality of the developer tools will be expanded and in the future there may be problems with the names of other scripts.

JulianOrteil:
@HeaTTheatR I have updated the console scripts to reflect the suggestions made by @Kulothungan16. As I don't have access to MacOS or easy access to Linux, I cannot vouch for the compatibility of this command theme with those OS's. They work fine on Windows 10. You'll want to verify those OS's work before merging this PR.

Issue #1327, opened 2022-08-30
MDRaisedButton - Elevation


arncode90:
### Description of the Bug

MDRaisedButton still showing an elevation (when the button is clicked), even I set the elevation to 0.

### Code and Logs

```python
from kivy.lang import Builder
from kivymd.app import MDApp

KV = '''
<RaisedBtn@MDRaisedButton>
    elevation: 0
    shadow_offset: (0,0)
    shadow_radius: 0
    shadow_softness: 0
    ripple_scale: .1
    
MDScreen:
    MDBoxLayout:
        md_bg_color: 1, 1, 0, 1
        size_hint: 0.9, 0.7
        orientation: "vertical"
        pos_hint: {'center_x':.5, 'y':0.1}

        ScreenManager:
            MiniScreen:
                name: "one"

            MiniScreen2:
                name: "two"

<MiniScreen@MDScreen>:
    MDBoxLayout:
        orientation: "vertical"
        MDWidget:
        MDBoxLayout:
            orientation: "vertical"
            spacing: dp(10)
            adaptive_height: True

            RaisedBtn:
                size_hint_x: 1
                text: "Login"
                on_press: 
                    root.parent.current = 'two'
        MDWidget:   
        
<MiniScreen2@MDScreen>:
    MDBoxLayout:
        orientation: "vertical"
        MDWidget:
        MDBoxLayout:
            orientation: "vertical"
            spacing: dp(10)
            adaptive_height: True

            RaisedBtn:
                size_hint_x: 1
                text: "Register"
                # on_press: print(f"{self.text}")
                on_press: root.parent.current = 'one'
        MDWidget:   
'''


class Test(MDApp):
    def build(self):
        return Builder.load_string(KV)


Test().run()
```

### Video
https://user-images.githubusercontent.com/48113758/187356140-a6b67545-3810-4cab-8378-cb513a6efeb4.mov

### Versions

* OS: Windows
* Python: v3.7.6
* Kivy: v2.1.0
* KivyMD: 1.1.0.dev0, git-Unknown, 2022-08-28

mp-007:
@arncode90  When you run your code. What it's your opengl version?
![image](https://user-images.githubusercontent.com/19823482/187700720-b02d8ac5-9cbe-4265-a48f-6717ccabdf30.png)

HeaTTheatR:
@mp-007 
[...]

PR #1331, opened 2022-09-02
fix elevation behavior when widget has Screen type as parent

mp-007:
possible fix for issue #1327

### DIFF
--- a/kivymd/data/glsl/elevation/elevation.frag
+++ b/kivymd/data/glsl/elevation/elevation.frag
@@ -10,6 +10,12 @@ corner radius:
 https://iquilezles.org/articles/distfunctions
 */
 
+float mysmoothstep(float a, float b, float x) {
+    float t = clamp((x - a) / (b - a), 0.0, 1.0);
+ 
+    return t * t * (3.0 - (2.0 * t));
+}
+
 float roundedBoxSDF(vec2 centerPosition, vec2 size, vec4 radius) {
     radius.xy = (centerPosition.x > 0.0) ? radius.xy : radius.zw;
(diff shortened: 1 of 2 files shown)


HeaTTheatR:
@mp-007 There is still a bug for this issue:


https://user-images.githubusercontent.com/16930280/188254721-aa9635b0-f416-42a7-b41f-b93065b3c052.mov

mp-007:
Maybe we can find something in this project: https://github.com/kivy-garden/frostedglass/blob/main/kivy_garden/frostedglass/__init__.py

I think the author was facing the same issues with the shader effect.

HeaTTheatR:
@mp-007 
Yes, it is quite obvious that the problem is in the shader itself. The problem of widgets with elevation behavior in scrollable lists has been solved. But the solution is not quite right. We knew that.

```python
from kivy.lang import Builder

from kivymd.app import MDApp


class Test(MDApp):

    def build(self):
        return Builder.load_string(
            '''
MDScreenManager:

    MDScreen:
        name: "screen 1"

        MDRaisedButton:
            text: "MDRaisedButton"
            on_release: root.current = "screen 2"

    MDScreen
        name: "screen 2"

        MDRaisedButton:
            text: "MDRaisedButton"
            on_release: root.current = "screen 3"

    MDScreen
        name: "screen 3"

        MDRaisedButton:
            text: "MDRaisedButton"
            on_release: root.current = "screen 1"
'''
        )


Test().run()
```

https://user-images.githubusercontent.com/16930280/188297360-36ce3fd8-b855-4a5b-987a-70fb81af41ad.mov


Unfortunately, I don't have too much knowledge on working with shaders.

mp-007:
PR #1336 fix issue #1327

PR #1342, opened 2022-09-07
Fix MDSegmentedControl segment_switch.width

inna-voig:
Fix segment_switch.width with a bind in the kv file


HeaTTheatR:
@inna-voig Good PR.

PR #1349, opened 2022-09-14
Update create_project.py

did0u007:
Fix Windows compatibility

First code was able to work perfectly on iOS and Linux but not on windows  because of paths and architectures are not the same  

Description of the algorithm of actions

Sans titre

### Description of Changes

I did handle Windows paths and, multiples pythons versions and taking care with venvs.

### DIFF
--- a/kivymd/tools/patterns/create_project.py
+++ b/kivymd/tools/patterns/create_project.py
@@ -657,13 +657,25 @@ def main():
     python_version = args.python_version
     if "3" not in python_version:
         parser.error("Python must be at least version 3")
+
+    # Fixed to work on Windows
+    if os.name == "nt":
+        project_directory = os.path.realpath(args.directory)
+        envs = [i.strip("\\") for i in os.getenv("PATH").split(";") if i != ""]
+        for i in envs:
+            env = [j.lower() for j in i.split("\\")]
+            if python_version.replace(".", "").lower() == env[-1]:
+                if os.path.exists(os.path.join(i, "python.exe")):
+                    python_version = os.path.join(i, "python.exe")
+                else:
+                    parser.error("Python version you entred must be installed")
+                break
+
     name_screen = args.name_screen
     path_to_project = os.path.join(project_directory, project_name)
     name_database = args.name_database
     if name_database != "no" and name_database not in available_databases:
-        parser.error(
-            f"The database name must be one of the {available_databases} list"
-        )
+        parser.error(f"The database name must be one of the {available_databases} list")
     use_hotreload = args.use_hotreload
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
@did0u007 We strictly adhere to the `PEP8` standards. In your changes, this standard is completely ignored. After making the changes, run the `pre-commit run --all-files` command to automatically format the code according to the `PEP8` standards.

kivymd/tools/patterns/create_project.py
```diff
@@ -657,13 +657,25 @@ def main():
...
+        envs = [i.strip("\\") for i in os.getenv("PATH").split(";") if i != ""]
+        for i in envs:
+            env = [j.lower() for j in i.split("\\")]
+            if python_version.replace(".", "").lower() == env[-1]:
+                if os.path.exists(os.path.join(i, "python.exe")):
+                    python_version = os.path.join(i, "python.exe")
+                else:
+                    parser.error("Python version you entred must be installed")
+                break
+
```
HeaTTheatR, line 673: I don't quite understand the meaning of your changes and how they relate to the error in the screenshot. Could you describe in more detail the problems that gets in OS Windows and how your changes solve these problems. Thanks.

did0u007: <sub>


    # Fixed to work on Windows
    if os.name == "nt":               # chek for Windows OS

        # get full path to project dir cuz this type of path "./path/to/project"  will cause problems if cmd get lunched from
        # different dir so path must be like this "c:/path/to/project"
        project_directory = os.path.realpath(args.directory)

        # get access to Variables Environment to check all pythons versions installed paths
        envs = [i.strip("\\") for i in os.getenv("PATH").split(";") if i != ""]


        for i in envs:
            env = [j.lower() for j in i.split("\\")]
            if python_version.replace(".", "").lower() == env[-1]: 
            
                 # in Windows Ex: python3.10 will be in "../Python310/python.exe " Dir and check if is really installed not only 
                 # an entry in OS Env list 
                if os.path.exists(os.path.join(i, "python.exe")):
                    python_version = os.path.join(i, "python.exe")
                else:
                    parser.error("Python version you entred must be installed")
                break
</sub>

HeaTTheatR:
```
    # get full path to project dir cuz this type of path "./path/to/project"  will cause problems if cmd get lunched from
    # different dir so path must be like this "c:/path/to/project"
```

The path to the project is specified by the user. And all responsibility for the correctness of the path to the project lies with the user.

HeaTTheatR:

PR #1357, opened 2022-09-24
PEP8-compliant KivyMD project autogen

JulianOrteil:

KivyMD's `create_project` generator is not PEP8 compliant with how it names its modules. For example, it creates `View`, `Model`, and `Controller` directories whereas PEP8 specifies these should all be lowercase. The generator also uses the supplied view name for the package the view is in (i.e. `View. ---> MainScreen <--- .main_screen`) which is also not correct.

Also found a couple of typos, incompatibility with Windows, and unclear usage of "camel case".

See:
https://peps.python.org/pep-0008/#package-and-module-names
https://cs50.harvard.edu/python/2022/psets/2/camel/
https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/x2dbyw72(v=vs.71)?redirectedfrom=MSDN
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/naming-rules#naming-style-properties

### Description of Changes

- Modified generators to use the plural, lower snake_case style for module/package names
- Fixed a couple of typos
- Clarified which type of camel case to use (replaced with PascalCase). While it is true that the style allows both upper and lower characters for the first word, it does not enforce either which can be confusing. PascalCase requires all words to be capitalized and is generally more recognized as the common term to mean this style.
- Added a documentation warning about incompatibilities with Windows. Fixing them is outside the scope of this PR, but can be addressed pretty easily.

image


HeaTTheatR:
Model/View/Controller are not modules but packages. Therefore, they are named correctly. Otherwise, I also do not find any deviations from the PEP8 standard.

JulianOrteil:
@HeaTTheatR Your assertion is incorrect.
image

JulianOrteil:
I actually need to make a change to remove the underscores from the package names. Did not know that was discouraged, so do not merge yet.

HeaTTheatR:
@JulianOrteil No. The module name must have an underscore. If there is a class named `ClassName` in this module, then the module should be named `class_name.py`

JulianOrteil:
@HeaTTheatR Correct, which is why I am removing the underscores from the package, not the module.

For example:
```from views.main_screen.main_screen import MainScreen```
to
```from views.mainscreen.main_screen import MainScreen```

HeaTTheatR:
@JulianOrteil Developer Tools `create_project.py` does not create packages with an underscore.

JulianOrteil:
@HeaTTheatR I had it do that in this PR. I'm revising the PR now so it doesn't add those underscores to package names.

HeaTTheatR:
@JulianOrteil Read carefully again - `Developer Tools create_project.py does not create packages with an underscore`.

HeaTTheatR:
@JulianOrteil At the moment, the module `create_project.py` does not create packages and underscores.

JulianOrteil:
@HeaTTheatR I know it doesn't on your master branch. But in _this PR_ I had `create_project.py` add those underscores. I'm revising _this PR_ to ensure _this PR's_ `create_project` doesn't do that.

Current `create_project` on `KivyMD/master`: `from View.MainScreen.main_screen import MainScreenView`
**This PR before my revisions:** `from views.main_screen.main_screen import MainScreenView`
**This PR after my revisions:** `from views.mainscreen.main_screen import MainScreenView`

JulianOrteil:
A fully PEP8 compliant project:

Note `views/mainscreen/main_screen.py` is no longer `views/main_screen/main_screen.py` as seen in the screenshot above.
image

Issue #1169, opened 2022-01-14
MDDatePicker: Switching to year selection and back messes up day selection


floi:
### Description of the Bug

Steps to reproduce:

In freshly opened instance of `MDDatepicker`:

1. Select a different day than the default (today).
2. Switch to year mode via the triangle icon.
3. Without changing the year, switch back to day selection.
4. Both the default/today and the day selected in step 1 are highlighted.
5. Switch to year mode via the triangle icon.
6. Without changing the year, switch back to day selection.
7. Only the default (today) is highlighted.

Saving via the OK button after either step 4 or step 7 results in the default date without accounting for the changed day.

I've looked at the code and my impression is that `MDDatePicker`'s attributes `day`, `month`, `year` and `sel_day`, `sel_month`, `sel_year` are not used consistently. The docstrings suggest that the ones without the `sel_` prefix are (only?) used as defaults before the user has made any selection, but [this line](https://github.com/kivymd/KivyMD/blob/1cb82f7d2437770f71be7c5a4f7de4b8da61f352/kivymd/uix/pickers/datepicker/datepicker.py#L890) uses `day` which causes this issue. Changing that to `sel_day` seems to fix it.

### Code and Logs

```python
from kivy.core.window import Window
from kivy.lang import Builder

from kivymd.app import MDApp
from kivymd.uix.pickers import MDDatePicker

KV = '''
MDFloatLayout:

    MDToolbar:
        title: "MDDatePicker"
        pos_hint: {"top": 1}
        elevation: 10

    MDRaisedButton:
        text: "Open time picker"
        pos_hint: {'center_x': .5, 'center_y': .5}
        on_release: app.show_date_picker()
'''


class Test(MDApp):
    def build(self):
        return Builder.load_string(KV)

    def on_save(self, instance, value, date_range):
        '''
        Events called when the "OK" dialog box button is clicked.

        :type instance: <kivymd.uix.picker.MDDatePicker object>;

        :param value: selected date;
        :type value: <class 'datetime.date'>;

        :param date_range: list of 'datetime.date' objects in the selected range;
        :type date_range: <class 'list'>;
        '''

        print(instance, value, date_range)

    def on_cancel(self, instance, value):
        '''Events called when the "CANCEL" dialog box button is clicked.'''

    def show_date_picker(self):
        date_dialog = MDDatePicker()
        date_dialog.bind(on_save=self.on_save, on_cancel=self.on_cancel)
        date_dialog.open()


Test().run()

```

[...]

PR #1370, opened 2022-10-03
Fix `MDDatePicker` bug (#1169)

nanouasyn:

See #1169. Switching to the year selection dialog and back resets the selected date to default.

### Description of Changes

@floi is right. `MDDatePicker.transformation_from_dialog_select_year` method (that is called when switching back via the triangle icon) uses the `day` field, which is responsible for the default date day set by the user, instead of using the `sel_day` field, which is responsible for the selected date day. I have fixed it.

In the same method, the author makes a similar mistake: he uses `year` and `month` instead of `sel_year` and `sel_month` to update the calendar. It works because the same mistake is made in the `DatePickerYearSelectableItem.on_touch_down` method, which, when clicked on the year, sets the `year` field instead of `sel_year`. It's not directly related to the bug, but I have fixed it in the same PR, because it's very close to the issue and because it threatens to become a source of bugs in the future.

In addition, an error in the `MDDatePicker.set_month_day` method caused the selection not to be removed from the previously selected date widget, and we saw two selected dates at the same time — the old and the new.

```python
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDFloatLayout:

    MDTopAppBar:
        title: "MDDatePicker"
        pos_hint: {"top": 1}
        elevation: 10

    MDRaisedButton:
        id:button
        text: "Open date picker"
        pos_hint: {'center_x': .5, 'center_y': .5}
        on_release: app.show_date_picker()
'''


class Test(MDApp):
    def build(self):

    def on_save(self, instance, value, date_range):
        '''
        Events called when the "OK" dialog box button is clicked.

        :type instance: <kivymd.uix.picker.MDDatePicker object>;

        :param value: selected date;
        :type value: <class 'datetime.date'>;

        :param date_range: list of 'datetime.date' objects in the selected range;
        :type date_range: <class 'list'>;
        '''

        print(instance, value, date_range)
        self.root.ids.button.text = str(value)

    def on_cancel(self, instance, value):
        '''Events called when the "CANCEL" dialog box button is clicked.'''

    def show_date_picker(self):
        self.date_dialog = MDDatePicker()
        self.date_dialog.bind(on_save=self.on_save, on_cancel=self.on_cancel)
        self.date_dialog.open()

Test().run()
```

Run the code:

image

Click on the button:

image

Select a different day than the default (October 20):

image

Switch to year mode via the triangle icon:

image

Without changing the year, switch back to day selection:

image

The date we selected earlier still seems selected. 

Click on the OK:

image

The date we selected is written on the button, as expected.

### DIFF
--- a/kivymd/uix/pickers/datepicker/datepicker.py
+++ b/kivymd/uix/pickers/datepicker/datepicker.py
@@ -771,7 +771,7 @@ class DatePickerYearSelectableItem(RecycleDataViewBehavior, MDLabel):
         if super().on_touch_down(touch):
             return True
         if self.collide_point(*touch.pos) and self.selectable:
-            self.owner.year = int(self.text)
+            self.owner.sel_year = int(self.text)
             # self.owner.sel_year = self.owner.year
             self.owner.ids.label_full_date.text = self.owner.set_text_full_date(
                 self.owner.sel_year,
@@ -1077,10 +1077,10 @@ class MDDatePicker(BaseDialogPicker):
 
         self._calendar_layout.clear_widgets()
         self.generate_list_widgets_days()
-        self.update_calendar(self.year, self.month)
+        self.update_calendar(self.sel_year, self.sel_month)
 
         if self.mode != "range":
-            self.set_month_day(self.day)
+            self.set_month_day(self.sel_day)
             self._sel_day_widget.dispatch("on_release")
 
     def transformation_to_dialog_select_year(self) -> None:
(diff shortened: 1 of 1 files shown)


nanouasyn:
I was wrong. In fact, `month` and `year` are not used as the default numbers of the month and year, but as the month and year of the displayed calendar page. 

When choosing a different year, we have to change the displayed page relative to the page currently displayed. However, my changes move the selected date relative to the previously selected one, after which they show the user a page with the selected date. Thus, if we switch to another month, then switch to the year selection and back, we will not go back, but will find ourselves on a page with a selected date.

On the other hand, the behavior before my correction also seems wrong. In this case, when we switch to another month, then switch to the year selection, and back, we will find ourselves in the month from which we switched to the year selection, but the selection will move to this month.

Ideally, we should be where we switched to the year selection from, and the selection should remain where it was. I need to think a little more about how to achieve this behavior. I will then add the commit to this branch.

nanouasyn:
I'm closing this PR because of #1375.

PR #1394, opened 2022-10-14
Add `on_copy` event when `on_double_tap` dispatch and `allow_copy` flag to `MDLabel`

Neizvestnyj:
```py
from kivymd.app import MDApp
from kivymd.uix.label import MDLabel

KV = """
<CopyLabel@MDLabel>
    halign: 'center'
    allow_copy: True
    md_bg_color: (0, 0.5, 0, 1)
    on_copy: print(f'Copied {self.text}')

ScrollView:    
    MDBoxLayout:
        padding: dp(10)
        spacing: dp(10)
        adaptive_height: True
        
        MDBoxLayout:
            orientation: 'vertical'
            padding: dp(10)
            spacing: dp(50)
            adaptive_height: True
            
            CopyLabel:
                text: 'Text to copy 1'
                adaptive_height: True
            
            CopyLabel:
                text: 'Text to copy 2'
                adaptive_height: True
            
            CopyLabel:
                text: 'Text to copy 3'
                adaptive_height: True
                
        CopyLabel:
            text: 'Text to copy 4'
        
"""


class TestApp(MDApp):
    def build(self):


TestApp().run()
```

### DIFF
--- a/kivymd/uix/label/label.py
+++ b/kivymd/uix/label/label.py
@@ -337,14 +337,6 @@ class MDLabel(
     and defaults to `None`.
     """
 
-    on_copy = ObjectProperty()
-    """
-    Label on copy event.
-
-    :attr:`on_copy` is an :class:`~kivy.properties.ObjectProperty`
-    and defaults to `None`.
-    """
-
     _text_color_str = StringProperty()
 
     parent_background = ColorProperty(None)
--- a/kivymd/uix/label/label.py
+++ b/kivymd/uix/label/label.py
@@ -379,8 +379,8 @@ class MDLabel(
         # TODO: Add letter spacing change
         # self.letter_spacing = font_info[3]
 
-    def on_long_touch(self, *args):
-        if self.allow_copy:
+    def on_double_tap(self, touch, *args):
+        if self.allow_copy and self.collide_point(*touch.pos):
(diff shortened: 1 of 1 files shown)


kivymd/uix/label/label.py
```diff
@@ -322,6 +323,22 @@ def _set_text(self, value):
...
 
+    allow_copy = BooleanProperty(False)
+    """
+    Label allow copy flag.
+
+    :attr:`allow_copy` is an :class:`~kivy.properties.BooleanProperty`
+    and defaults to `None`.
+    """
+
+    on_copy = ObjectProperty()
```
HeaTTheatR: This is bug!

Neizvestnyj: Of course)

HeaTTheatR:
@Neizvestnyj Maybe it's better to use a double click instead of a long touch?

Neizvestnyj:
> @Neizvestnyj Maybe it's better to use a double click instead of a long touch?

I dont think so, users alway pinch the widget before copying

HeaTTheatR:
> > @Neizvestnyj Maybe it's better to use a double click instead of a long touch?
> 
> I dont think so, users alway pinch the widget before copying

Double-clicking on any system highlights the text. Not a long press.

PR #1407, opened 2022-11-01
added screen name property to generated .kv files in MVC project creator module

kengoon:
this PR adds the  screen`name` property to the generated .kv file. it complements the screen names on the `screen.py`

e.g
```kvlang
<NameScreenView>:
    name: 'name screen'
```


HeaTTheatR:
In the MVC template, this works differently. The names of the screens are contained in the file `View/screens.py`.

PR #1409, opened 2022-11-04
Add documentation for dialog's content_cls

Bsenss6:
The `content_cls` attribute of the dialog component is only used when
the `type` attribute is set to `'custom'`. This was not specified in
the documentation. I spent some time to understand that the `type` attribute was the issue. I hope this change will help others who will try to use this attribute.

This is my first pull request to this project, please let me know if you would prefer I made something different.

Thanks!

### DIFF
--- a/kivymd/uix/dialog/dialog.py
+++ b/kivymd/uix/dialog/dialog.py
@@ -563,7 +563,8 @@ class MDDialog(BaseDialog):
         :align: center
 
     :attr:`content_cls` is an :class:`~kivy.properties.ObjectProperty`
-    and defaults to `'None'`.
+    and defaults to `'None'`. This attribute is only available when
+    :attr:`type` is set to `'custom'`.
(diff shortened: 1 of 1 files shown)


kivymd/uix/dialog/dialog.py
```diff
@@ -563,7 +563,8 @@ def show_confirmation_dialog(self, *args):
         :align: center
 
     :attr:`content_cls` is an :class:`~kivy.properties.ObjectProperty`
-    and defaults to `'None'`.
+    and defaults to `'None'`. This attribute is only available when
```
HeaTTheatR: The description of the property should be at the beginning of the documentation line:

```python
    property = ObjectProperty()
    """
    Description.

    :attr:`property` is an :class:`~kivy.properties. ObjectProperty `
    and defaults to `None`.
    """
```

PR #1413, opened 2022-11-13
WIP Added `bar_radius` to `MDScrollView` class

Neizvestnyj:
Added a radius for the widget and put everything in a separate folder

```py
from kivymd.app import MDApp
from kivymd.uix.list import OneLineListItem

KV = """
Screen:
    MDScrollView:
        bar_color: [1, 0, 0, 1]
        bar_radius: [5, 5, 5, 5]
        scroll_type: ['bars', 'content']
        bar_width: 10
        MDList:
            id: container
            spacing: 2
            
"""


class TestApp(MDApp):
    def build(self):

    def on_start(self):
        for i in range(20):
            self.root.ids.container.add_widget(
                OneLineListItem(text=f"Single-line item {i}")
            )


TestApp().run()
```

image


Neizvestnyj:
I think everything is working correctly now

HeaTTheatR:
@Neizvestnyj We will not accept new pull requests yet. We need to cover the KivyMD library with tests. Otherwise, each new change causes a lot of bugs in other modules and widgets of the library.

Neizvestnyj:
> @Neizvestnyj We will not accept new pull requests yet. We need to cover the KivyMD library with tests. Otherwise, each new change causes a lot of bugs in other modules and widgets of the library.

OK

Neizvestnyj:
While no tests, I did not update this PR, current version works incorrect, I already have fixes, but they need to be discussed.

HeaTTheatR:

PR #1456, opened 2023-01-26
Expansion panel not closing

### LINKED ISSUE #1453: Expansion panel content.height
### Description of the Bug

I have a MDExpansionPanel in my app. On the very first click to expand the panel, the expansion panel grossly over extends the height. Closing the panel then re opening it the second time and it displays as i would expect it to.

Unfortunately i have tried to make a minimal code of the issue with no success. So this bug report is perhaps something to keep in the back of the mind in case it sparks a thought in the future.

**What i've noticed the cause is **
In expansionpanel.py and in the def open_panel() method there is a animation that uses content.height
for me the self.content.height is way larger than the desired figure on the second opening 
```
print(f"Open panel method, animation height -> \n content.height = {self.content.height} \n self.height = {self.height}")
anim = Animation(
            height=self.content.height + self.height,
            d=self.opening_time,
            t=self.opening_transition,
        )
        anim.bind(on_complete=self._add_content)
        anim.bind(on_complete=self._disable_anim)
```
When i add a print statement above that animation, i notice that on first click of the panel my self.content.height is 1690
On second click (when it displays correctly) the self.content.height is 560.
```
#first opening
Open panel method, animation height -> 
 content.height = 1690.0 
self.height = 56.0

# Second opening
Open panel method, animation height -> 
 content.height = 560 
self.height = 56.0

```
My panel Content is made up of a MDCard and around 6 MDLabels, but as i say i can't replicate it in minimal code just yet so i'm documenting it for reference  

### Versions

* OS: 
* Python: 
* Kivy: 
* KivyMD:

LazzaAU:

As documented by #1453 the expansion panel would not close using the close_panel method

### Description of Changes

There are required steps in ```check_open_panel``` method such as removal of the widget that doesn't get called in the ```on_close``` method. So i have refactored the ```on_close``` method to ```on_close_animation``` then made a new  ```on_close``` method that points directly to ```check_open_panel``` which is the existing method that prepares the expansion panel for opening and closing.

I'm not certain there is a need for the existing ```press_current_panel``` boolean, as i'm not sure when a user would choose to make that "False" ? however i've left it in there as its existing code  

If this PR is acceptable it will close #1453 

PS: Looks like my pycharm has formatted some other lines, hence the 54 additions :)

### DIFF
--- a/.idea/KivyMD.iml
+++ b/.idea/KivyMD.iml
@@ -2,6 +2,7 @@
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$" isTestSource="false" />
       <excludeFolder url="file://$MODULE_DIR$/venv" />
     </content>
     <orderEntry type="inheritedJdk" />
--- a/kivymd/uix/button/button.py
+++ b/kivymd/uix/button/button.py
@@ -2211,11 +2211,6 @@ class MDFloatingActionButtonSpeedDial(
         def set_pos_root_button(*args):
             if self.anchor == "right":
                 instance_floating_root_button.y = dp(20)
-                self.startupcheck = False
-                print(f"Lazza - speedDial Button parent width is {self.parent.width} is it less than 130 ?")
-                if not self.startupcheck and self.parent.width < 130:
-                    self.parent.width = 430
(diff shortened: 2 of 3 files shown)


HeaTTheatR:
I think this is not a very good solution that cannot be merged with the main branch of the KivyMD library.

LazzaAU:
No worries :)

Issue #1457, opened 2023-01-29
MDExpansionPanel doesn't work well in MDBottomNavigationItem


PythonLaboratory:
### Description of the Bug

MDExpansionPanel does not work well in MDBottomNavigationItem: it does not open and close well or even flies away. If MDBottomNavigationItem is not used, then everything is fine.

### Code and Logs

```python
import os


from kivy.lang import Builder

from kivymd.app import MDApp
from kivymd.uix.tab import MDTabsBase
from kivymd.uix.floatlayout import MDFloatLayout
from kivymd.uix.boxlayout import MDBoxLayout
from kivymd.icon_definitions import md_icons
from kivymd.uix.expansionpanel import MDExpansionPanel, MDExpansionPanelThreeLine
from kivymd import images_path


KV = '''
MDBoxLayout:
    orientation: "vertical"

    MDTopAppBar:
        title: "Example Tabs"

    MDBottomNavigation:

        MDBottomNavigationItem:
            text: 'home'
            icon: 'home'
            
            MDTabs:
                id: tabs

<Content>
    adaptive_height: True

    TwoLineIconListItem:
        text: "(050)-123-45-67"
        secondary_text: "Mobile"

        IconLeftWidget:
            icon: 'phone'

<Tab>
    orientation: 'vertical'

    MDScrollView:

        MDGridLayout:
            id: box
            cols: 1
            adaptive_height: True
'''


class Content(MDBoxLayout):
    '''Custom content.'''


class Tab(MDBoxLayout, MDTabsBase):
    '''Class implementing content for a tab.'''

    def on_kv_post(self, *args):

        self.load()
        
    def load(self):

        for i in range(3):
            self.ids.box.add_widget(
                MDExpansionPanel(
                    icon=os.path.join(images_path, "logo", "kivymd-icon-128.png"),
                    content=Content(),
                    panel_cls=MDExpansionPanelThreeLine(
                        text="Text",
                        secondary_text="Secondary text",
                        tertiary_text="Tertiary text",
                    )
                )
            )

            
class Example(MDApp):
    icons = list(md_icons.keys())[15:30]

    def build(self):
        
        self.theme_cls.theme_style = "Dark"
        self.theme_cls.primary_palette = "Orange"
        
        return Builder.load_string(KV)

    def on_start(self):
        for tab_name in self.icons:
            self.root.ids.tabs.add_widget(Tab(icon=tab_name))


Example().run()
```

How to fix it right now?

### Versions

[...]

PR #1458, opened 2023-01-31
Fixed using MDExpansionPanel inside MDBottomNavigation

PythonLaboratory:

MDExpansionPanel does not work well in MDBottomNavigationItem: it does not open and close well or even flies away. [#1457](url)

### Description of Changes

Now `check_open_panel` is called on the `on_press` event, not `on_release`


PythonLaboratory:
Maybe this is not the best solution, but at least it solves the problem

HeaTTheatR:
I think this is not a very good solution that cannot be merged with the main branch of the KivyMD library - https://github.com/kivymd/KivyMD/pull/1456

Issue #1441, opened 2022-12-26
ZeroDivisionError on exiting the header of a sortable column


TMaddox:
### Description of the Bug

On executing the code from https://kivymd.readthedocs.io/en/1.1.1/components/datatables/ I get the following error:

```
Exception ignored in: 'kivy.graphics.vertex_instructions.Line.prebuild_rounded_rectangle'
 Traceback (most recent call last):
   File "C:\Users\michael\miniconda3\envs\hamster\lib\site-packages\kivy\core\window\__init__.py", line 1627, in on_draw     self.render_context.draw()
 ZeroDivisionError: float division by zero
 ZeroDivisionError: float division by zero
```

The error seems to occur only when the header of a sortable column is exited (Signal Name, Schedule, Team Lead on the example from the docs). Also, the error does not crash the application.

### Code and Logs

https://kivymd.readthedocs.io/en/1.1.1/components/datatables/#api-kivymd-uix-datatables-datatables

### Versions

* OS: Windows 10
* Python: 3.9.15
* Kivy: 2.1.0
* KivyMD: 1.1.1

Obliman:
I found that this error can be prevented by changing the following line in `kivymd/uix/tooltip/tooltip.py`:

`anim = Animation(_scale_x=0, _scale_y=0, d=0.1) + Animation(opacity=0, d=0.2)`

Becomes:

`anim = Animation(_scale_x=1, _scale_y=1, d=0.1) + Animation(opacity=0, d=0.2)`

The difference is setting `_scale_x` and ` _scale_y` to 1 instead of 0. Visually I don't see a difference, so I'm not clear on what setting zero scale values does. My guess is that calculating a radius for a zero dimension is what leads to the error...?

oliverbrace:
Not sure when people think this will be sorted but is anyone aware of a work around for now? 
Thanks

HeaTTheatR:
Fixed in master branch.

PR #1475, opened 2023-03-17
Address header sort button divide by zero error

nhansendev:

Addresses the bug mentioned in #1441 by removing the lines that resize the sort button to have a height of zero, which leads to a divide by zero error when calculating rounded rectangle radius. Visually it looks the same as before the error was fixed since the items being hidden also have opacity toggled as needed.

See the example code in the [documentation for datatables](https://kivymd.readthedocs.io/en/1.1.1/components/datatables/). Zero-division errors only appeared when the mouse moved out of one of the **sortable** header titles, which causes the sort button to be resized while disappearing.

### Reproducing the problem

See the example code in the [documentation for datatables](https://kivymd.readthedocs.io/en/1.1.1/components/datatables/).

kivymd_bug_1441_

### Description of Changes

Removing the lines that resize the button to have a height of zero.

See the example code in the [documentation for datatables](https://kivymd.readthedocs.io/en/1.1.1/components/datatables/).


nhansendev:
Note: as far as I know setting the button size to zero serves no purpose as long as setting opacity to zero is sufficient. If there is a situation where opacity leaves something still visible, or the size of the header is affected by the sort button size not being zero, then we would need a different solution. Surely there is a more "proper" way to hide a widget in that case?

HeaTTheatR:
@Obliman In this case, we need to set the button to the disabled/disabled state, because having zero `opacity`, the sort button will still respond to user events.

nhansendev:
Added the button enable/disable lines alongside the opacity ones.

HeaTTheatR:

PR #1482, opened 2023-04-07
Time picker format

edwinjosegeorge:

While setting time, if the hour is 0, then it should convert to 12 AM

call set_time(datetime.time(hour=0, minute=22))

### Reproducing the problem

```python
my_time = MDTimePicker()
my_time.set_time(datetime.time(hour=0, minute=20))
```

image

### Description of Changes

Use of standard library for conversion

image

```python
my_time = MDTimePicker()
my_time.set_time(datetime.time(hour=0, minute=20))
```


HeaTTheatR:

PR #1483, opened 2023-04-10
Fix-Pillow-Update-Version

aamirawan7584:

Kivy MD internally uses Pillow for drawings and it was working fine.
Pillow team has updated the pillow to version 9.5.0 which seems to be not compatible with Kivy MD so until we change implementation we can specify the previous Pillow stable version for installation.
[Pillow Release Notes:](https://pillow.readthedocs.io/en/stable/releasenotes/9.5.0.html)

Description of the algorithm of actions

### Reproducing the problem
Just try to run any kivy md application without updating the pillow version.

kivymdIssue

### Description of Changes

Fixed Pillow version to 9.4.0 which is a previous stable version of Pillow and works well with Kivy MD


HeaTTheatR:
We don't use Pillow for elevation behavior.

PR #1513, opened 2023-06-18
better touch behaviour

Kak0ytachel:

unable to bind outer `on_long_touch` method (and simmilar ones)

calling on_long_touch() itself whithout dispatching

### Reproducing the problem

```python
from kivy.uix.button import Button
from kivymd.app import MDApp
from kivymd.uix.behaviors import TouchBehavior


class MainApp(MDApp):
    def build(self):
        btn = MyButton()
        btn.text = 'Hello world'
        btn.bind(on_long_touch=self.long_touch)
        return btn

    def long_touch(self, *args):
        print('outer long touch')


class MyButton(Button, TouchBehavior):
    def on_long_touch(self, touch, *args):
        print('inner long touch')


MainApp().run()
```

### Description of Changes

calling methods via dispatcher

### DIFF
--- a/kivymd/uix/behaviors/touch_behavior.py
+++ b/kivymd/uix/behaviors/touch_behavior.py
@@ -68,6 +68,9 @@ class TouchBehavior:
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.register_event_type("on_long_touch")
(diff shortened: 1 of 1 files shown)


HeaTTheatR:

PR #1546, opened 2023-08-24
Allow screens with multiple uppercase words to be added to screens.py

krille90:
allow to add screen with multiple UpperCase words, e.g.  "NotificationSettingsScreen" to the screens.py file

Currently, when you add a view using kivymd.add_view, screens with more than two words in name won't be added properly to screens.py and need to added manually.

The problem lies in the add_view.py, which only allows name length of 2 (`len(res) == 2`)

### Reproducing the problem

```sh
kivymd.add_view MVC . DebugMenuScreen
```

### Description of Changes

the check was altered, such that it now allows screens with the minimum length of two words (`len(res) >= 2`)
Additionally the screen list key generation was changed to use all words in the screen name

```sh
kivymd.add_view MVC . DebugMenuScreen
```


HeaTTheatR:

PR #1580, opened 2023-12-23
Use `mActivity` provided by `android`

T-Dynamos:
Because `org.kivy.android.PythonActivity` is not same in every app and `android` module selects name itself. 

See:
https://github.com/kivy/buildozer/blob/079101d9086846c3fc5066de40506742bc044c9a/buildozer/default.spec#L145-L153


HeaTTheatR:
I think you should have made a pull request to the `2.0.0` branch, not the `1.2.0` branch.

T-Dynamos:
yes, I know.

PR #1591, opened 2024-01-21
Fix docs build v2

p0lygun:
Description of the problem
The docs CI is failing as we import kivymd in docs generation.

Describe the algorithm of actions that leads to the problem
This happens because the version of the package is defined in the __init__.py of kivymd.
When importing that it also runs [this](https://github.com/kivymd/KivyMD/blob/master/kivymd/__init__.py#L64) line, which then imports kivy, this issue should be fixed in the latest kivy, but its best to have version info in a separate file


HeaTTheatR:
Is this PR ready to merge?

p0lygun:
yes, all build system should work the same, but if anything breaks please feel free to ping me

HeaTTheatR:
Okay, thanks

D111sP:
Hello everyone, the large Materialyoucolor library does not want to work on a 32-bit system

D111sP:
ImportError: dlopen failed: "/data/data/com.autograd.ag_emp/files/app/_python_bundle/site-packages/materialyoucolor/quantize/celebi.so" is 64-bit instead of 32-bit

HeaTTheatR:
Are 32-bit systems still supported?

T-Dynamos:
Did you use recipe for building?

> ImportError: dlopen failed: "/data/data/com.autograd.ag_emp/files/app/_python_bundle/site-packages/materialyoucolor/quantize/celebi.so" is 64-bit instead of 32-bit

p0lygun:
why is this discussion taking place in the PR ?

PR #1629, opened 2024-02-27
(`uix/scrollview`): Implement material design specifications for `MDS…

### LINKED ISSUE #1602: Implement new "stretch and bounce back" scroll effect
With Android 12:

| Pull stretch | Fling stretch |
| ------------- | ------------- |
| image  | image |

> Starting in Android 12 (API level 31), the visual elements stretch and bounce back on a drag event and fling and bounce back on a fling event.

Source: https://developer.android.com/develop/ui/views/touch-and-input/gestures/scroll

T-Dynamos:
Closes #1602

### DIFF
--- a/examples/material_scroll.py
+++ b/examples/material_scroll.py
@@ -1,0 +1,124 @@
+import os
+import sys
+
+from kivy.core.window import Window
+from kivy.metrics import dp
+from kivy.uix.boxlayout import BoxLayout
+from kivy.lang import Builder
+from kivy import __version__ as kv__version__
+from kivymd import __version__
+from kivymd.app import MDApp
+from materialyoucolor import __version__ as mc__version__
+
+from examples.common_app import CommonApp, KV
+
+MAIN_KV = """
+<Item>:
+    size_hint_y:None
+    height:dp(50)
+    text:""
+    sub_text:""
+    icon:""
+    spacing:dp(5)
+    MDIcon:
+        icon:root.icon
+        size_hint:None, 1
+        width:self.height
+    BoxLayout:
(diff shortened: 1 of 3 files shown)


main.py
HeaTTheatR, line 1: Can you delete this file? It is not needed in the modules of the KivyMD library. It would be better if we add this file to the examples folder - `examples`

T-Dynamos: @HeaTTheatR Done, Please test for further bugs.

HeaTTheatR:
In this implementation, the behavior needs to be slightly changed. In Android, it behaves somewhat differently, as far as I know. At the moment, the zoom effect is applied at the moment when the list has reached the end/beginning. In Android, the scale is also applied at the over_scroll event.

HeaTTheatR:
I can't help with this right now, because there are already a lot of changes in my local repository related to dynamic color, which does not work with the Python implementation. That is, I still have a lot of work to finish.

T-Dynamos:
> In this implementation, the behavior needs to be slightly changed. In Android, it behaves somewhat differently, as far as I know. At the moment, the zoom effect is applied at the moment when the list has reached the end/beginning. In Android, the scale is also applied at the over_scroll event.

You should try to force overscroll, it will zoom.

HeaTTheatR:
Cool! It's working now! Great job!

examples/material_scroll.py
HeaTTheatR, line 1: `ModuleNotFoundError: No module named 'psutil'`

In `Unix`, as I understand it, there is no pre-installed module `psutil`

Issue #1644, opened 2024-03-12
Removing MDListItemTrailingIcon raise an exception


ikus060:
### Description of the Bug

When trying to remove the MDListItemTrailingIcon from a MDListItem an exception is raised.

### Code and Logs

```python
from kivy.lang import Builder
from kivy.uix.boxlayout import BoxLayout
from kivymd.app import MDApp

KV = '''
<MyView>:

    MDBoxLayout:
        orientation: 'vertical'

        Button:
            text: "Toggle icon"
            on_release: root.toggle()

        MDListItem:
            id: item

            MDListItemHeadlineText:
                text: "Headline"

            MDListItemTrailingIcon:
                id: trailing_icon
                icon: "trash-can-outline"

MDScreen:

    MDBoxLayout:
        id: content

        MyView:
'''

class MyView(BoxLayout):
    _trailing_icon = None

    def toggle(self):
        if self._trailing_icon:
            self.ids.item.add_widget(self._trailing_icon)
        else:
            self._trailing_icon = self.ids.trailing_icon
            self._trailing_icon.parent.remove_widget(self._trailing_icon)

class Example(MDApp):

    def build(self):
        return Builder.load_string(KV)
    
Example().run()

"""


class MainApp(App):
    def build(self):
        self.root = Builder.load_string(kv)


if __name__ == '__main__':
    MainApp().run()
```

Here the exception getting raised.
```
...
   File "/home/ikus060/Downloads/test_kivy.py", line 41, in toggle
     self._trailing_icon.parent.remove_widget(self._trailing_icon)
   File "/home/ikus060/workspace/PDSL/minarca.git/.venv/lib/python3.10/site-packages/kivy/uix/boxlayout.py", line 331, in remove_widget
     return super(BoxLayout, self).remove_widget(widget, *args, **kwargs)
   File "/home/ikus060/workspace/PDSL/minarca.git/.venv/lib/python3.10/site-packages/kivy/uix/layout.py", line 105, in remove_widget
     super(Layout, self).remove_widget(widget, *args, **kwargs)
   File "/home/ikus060/workspace/PDSL/minarca.git/.venv/lib/python3.10/site-packages/kivy/uix/widget.py", line 714, in remove_widget
     self.children.remove(widget)
   File "kivy/properties.pyx", line 855, in kivy.properties.ObservableList.remove
   File "kivy/properties.pyx", line 806, in kivy.properties.observable_list_dispatch
[...]

PR #1646, opened 2024-03-12
Support remove_widget() on MDListItem

ikus060:

When trying to remove the MDListItemTrailingIcon from a MDListItem an exception is raised.
See #1644.

See #1644.

### Description of Changes

Add extra condition to avoid raising KeyError.

I did not write unit test since I did not see any testing done for MDList in general.

N/A

N/A

### DIFF
--- a/kivymd/uix/list/list.kv
+++ b/kivymd/uix/list/list.kv
@@ -76,7 +76,7 @@
         size_hint_x: None
         width: 0
         on_children:
-            if text_container.children: \
+            if text_container.children and self.children: \
             self.children[0].pos_hint = {"top": 1} \
             if len(text_container.children) == 3 else {"center_y": .5}
 


HeaTTheatR:
Thanks. Also note that the bug is in two places:

https://github.com/kivymd/KivyMD/blob/master/kivymd/uix/list/list.kv#L69
https://github.com/kivymd/KivyMD/blob/master/kivymd/uix/list/list.kv#L78

ikus060:
@HeaTTheatR 
I don't see any problem on line 69. The condition is working fine with or without children.

HeaTTheatR:
> @HeaTTheatR I don't see any problem on line 69. The condition is working fine with or without children.

```python
<MDListItem>
    ...

    BoxLayout:
        ...

    AnchorLayout:
        ...

        BoxLayout:
            ...

            on_children:
                if leading_container.children: \
                leading_container.children[0].pos_hint = {"top": 1} \
                if len(args[1]) == 3 else {"center_y": .5}

    BoxLayout:
        ...

        on_children:
            if text_container.children: \
            self.children[0].pos_hint = {"top": 1} \
            if len(text_container.children) == 3 else {"center_y": .5}
```

ikus060:
@HeaTTheatR Sorry, but you will need to be more explicit if you think there is an issue with line 69.

According to me the current condition on line 69 `if leading_container.children` will not lead to an exception if the container has or has no children.

On the contrary, the `trailing_container` boxlayout has an issue, because the `on_children` rule make reference to `self.children[0]` which is not checked upon.

PR #1647, opened 2024-03-12
Support remove_widget() on MDSwitch

ikus060:

Similar to #1646, when trying to clear MDSwitch widget an exception is raised.

Calling clear_widgets() on MDSwitch raises an exception.

### Reproducing the problem

```python
from kivy.lang import Builder
from kivy.uix.boxlayout import BoxLayout
from kivymd.app import MDApp

KV = '''
<MyView>:

    MDBoxLayout:
        orientation: 'vertical'

        Button:
            text: "Remove icon"
            on_release: root.remove()

        MDSwitch:
            id: switch
            active: True
            icon_active: "check"

MDScreen:

    MDBoxLayout:
        id: content

        MyView:
'''

class MyView(BoxLayout):
    _trailing_icon = None

    def remove(self):
        self.ids.switch.clear_widgets()
        # OR
        #thumb = self.ids.switch.ids.thumb
        #thumb.parent.remove_widget(thumb)

class Example(MDApp):

    def build(self):
    
Example().run()

```

N/A

### Description of Changes

New condition added to avoid ` AttributeError: 'NoneType' object has no attribute 'disabled'`

N/A

Same code below.

### DIFF
--- a/kivymd/uix/selectioncontrol/selectioncontrol.kv
+++ b/kivymd/uix/selectioncontrol/selectioncontrol.kv
@@ -58,6 +58,7 @@
         text_color:
             ( \
             ( \
+            ( \
             root.parent.icon_active_color \
             if root.parent.icon_active_color \
             else self.theme_cls.onPrimaryContainerColor \
@@ -76,7 +77,9 @@
             if root.parent.icon_active else \
             self.theme_cls.surfaceContainerHighestColor[:-1] \
             + [root.parent.switch_opacity_value_disabled_icon] \
-            )
+            ) \
+            ) \
+            if root.parent else self.theme_cls.transparentColor
 
 
 <MDSwitch>


kivymd/uix/selectioncontrol/selectioncontrol.kv
```diff
@@ -56,6 +56,8 @@
         pos_hint: {"center_x": .5, "center_y": .5}
         theme_text_color: "Custom"
         text_color:
+            self.theme_cls.surfaceContainerHighestColor if not root.parent else \
```
HeaTTheatR: I would solve this in the following way:

```python
<Thumb>
    [...]

    ThumbIcon:
        [...]

        text_color:
            [...]

            if root.parent else self.theme_cls.transparentColor
```

ikus060: Ok

Thanks for the recommendation. I pushed the modification.

ikus060:
Hello @HeaTTheatR 

I've made the change to the PR. Anything else need my attention to get this PR merged ?

PR #1724, opened 2024-07-29
Update expansionpanel.py

Sahil-pixel:
def _set_content_height(self, *args):
        self._original_content_height = self._content.height - dp(88)
        self._content.height = 0
IN THIS LINE self._original_content_height = self._content.height - dp(88) 
dp(88) subtraction is not required .  
If we add widgets in MDExpansionPanelContent widget then there is some error for calculation and MDExpansionPanelContent's children not set in boxlayout properly ,

### DIFF
--- a/kivymd/uix/expansionpanel/expansionpanel.py
+++ b/kivymd/uix/expansionpanel/expansionpanel.py
@@ -525,5 +525,5 @@ class MDExpansionPanel(DeclarativeBehavior, BoxLayout):
             return super().add_widget(widget)
 
     def _set_content_height(self, *args):
-        self._original_content_height = self._content.height - dp(88)
+        self._original_content_height = self._content.height 
         self._content.height = 0


Sahil-pixel:
### Also see panel._original_content_height is updated after some frame . so please check it . properly . MDExpansionPanel need to be fixed . see i use Clock to get updated  value panel._original_content_height. in Desktop it not needed but in mobile device it is needed . please see those things properly . 
```
def tap_expansion_chevron(self,panel,content,header,chevron):
		Animation(
     		padding=[0, dp(12), 0, dp(12)]
            if not panel.is_open
            else [0, 0, 0, 0],
            d=0.2,
       	 	).start(panel)
		
		
		print(panel._content)#=content
		print(panel._header) #=header
		print(panel._original_content_height)
		#panel._header = ObjectProperty()  # MDExpansionPanelHeader object
    	#_content = 
		#panel.open() if not panel.is_open else panel.close()
		Clock.schedule_once(partial(self._call,panel))


        
def _call(self,panel,dt):
		
          print(panel._content)#=content
          print(panel._header) #=header
          print(panel._original_content_height)
          panel.open() if not panel.is_open else panel.close()

```

PR #1728, opened 2024-08-15
Allow to add Views with more than one word with `kivymd.add_view` function

krille90:

Currently when creating views via `kivymd.add_view`, only views with one word, e.g. `NotificationScreen`, will be added correctly. When you try to add a view with the name `NotificationSettingsScreen`, all classes in the folders Model, Controller and View will be added. However, it won't be added in the `screens.py`(and no error / warning is thrown).

In the file `kivymd/tools/patterns/add_view.py` (line 158ff) there is a restriction for only allowing one word screens:

```python
if res and len(res) == 2 and res[-1] == "Screen":
                    screens += (
                        "\n    '%s': {"
                        "\n        'model': %s,"
                        "\n        'controller': %s,"
                        "\n    },"
                        % (
                            f"{res[0].lower()} {res[1].lower()}",
                            f'{"".join(res)}Model',
                            f'{"".join(res)}Controller',
                        )
...
```

### Reproducing the problem

First create an kivymd MVC project. Then type in the following in commandline:

```commandline
> kivymd.add_view MVC . NotificationSettingsScreen 
```

In the end, have a look into the `screens.py`, the imports are there, but not added in the `screens` dict.

Here you can see, that  NotificationSettingsScreen wasn't added to the `screens` dict in `screens.py`
image

### Description of Changes

I changed above code to allow multiple words:

```python
if res and len(res) >= 2 and res[-1] == "Screen":
                    screens += (
                        "\n    '%s': {"
                        "\n        'model': %s,"
                        "\n        'controller': %s,"
                        "\n    },"
                        % (
                            f'{" ".join(res).lower()}',
                            f'{"".join(res)}Model',
                            f'{"".join(res)}Controller',
                        )
...
```

Here you can see, that it is now added properly.
image

```commandline
> kivymd.add_view MVC . NotificationSettingsScreen 
```

### DIFF
--- a/kivymd/tools/patterns/add_view.py
+++ b/kivymd/tools/patterns/add_view.py
@@ -155,14 +155,14 @@ def create_screens_data(
         for name in os.listdir(path_to_view):
             if os.path.isdir(os.path.join(path_to_view, name)):
                 res = re.findall("[A-Z][a-z]*", name)
-                if res and len(res) == 2 and res[-1] == "Screen":
+                if res and len(res) >= 2 and res[-1] == "Screen":
                     screens += (
                         "\n    '%s': {"
                         "\n        'model': %s,"
                         "\n        'controller': %s,"
                         "\n    },"
                         % (
-                            f"{res[0].lower()} {res[1].lower()}",
+                            f'{" ".join(res).lower()}',
                             f'{"".join(res)}Model',
                             f'{"".join(res)}Controller',
                         )


HeaTTheatR:
This module needs to be completely rewritten in accordance with the new API version 2.0.0

krille90:
I understand, although I don't seem to have any issue with using this module using 2.0.x.
I think merging wouldn't be much of a problem though as it's quite the small change and a rewrite of the module will still take some time.

Issue #1722, opened 2024-07-24
`setup.py` omits assets when `kivymd` is in directory path


charliermarsh:
If you build `kivymd` from a directory that has `kivymd` in the path, the `.kv` and other asset files are omitted, as `out_files.append(filepath.split(f"kivymd{os.sep}")[1])` (in `setup.py`) truncates at the first instance of `kivymd`.

For example: running `python -m build` from `~/workspace/kivymd/foo` creates `kivymd-2.0.1.dev0.tar.gz` without any `.kv` files in it. While running from `~/workspace/bar/foo/` includes the `.kv` files.

Zwicii:
Are there any updates on this issue regarding the missing assets in setup.py? We're still encountering this when using uv to install KivyMD

PR #1730, opened 2024-08-30
Fix glob_paths to correctly include .kv files in package

koala-sloth:
- Modified glob_paths function to properly identify files with given extensions
- Now correctly traverses subdirectories and returns relative file paths
- Fixes issue where .kv files were missing from pip installations
- Also affects .pot and .po files, ensuring all necessary files are included

Description of the problem:
The glob_paths function in setup.py was not correctly identifying .kv files (and potentially .pot and .po files) in the KivyMD project structure. This led to these files being excluded from the built package, causing issues for users who installed KivyMD via pip.

Algorithm of actions that leads to the problem:

```
Run python3 setup.py
The glob_paths function is called to find .kv files
The function returns a list of directory names instead of file paths
The setup process doesn't include the .kv files in the package
```

Reproducing the problem:
```
# In setup.py, add this line before the setup() call:
print(f"kv files: {glob_paths('.kv')}")

# Then run:
python3 setup.py

# This outputs:
# kv files: ['KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/', 'KivyMD/']
```

Screenshots of the problem:
This is text-only, part of building python package, screenshots not needed.

Description of Changes:
We modified the glob_paths function in setup.py to correctly identify files with the given extension in all subdirectories of the 'kivymd' folder. The key changes were:

```
Changed the condition to check for files (not directories) that end with the given pattern.
Used os.path.relpath() to get the path relative to the 'kivymd' directory.
Simplified error handling to catch any ValueError that might occur.
```

Code for testing new changes:
```
# In setup.py, add this line before the setup() call:
print(f"kv files: {glob_paths('.kv')}")

# Then run:
python3 setup.py

# This should now output a list of actual .kv file paths, like:
# kv files: ['uix/imagelist/imagelist.kv', 'uix/appbar/appbar.kv', ...]
```

### DIFF
--- a/kivymd/_version.py
+++ b/kivymd/_version.py
@@ -1,5 +1,5 @@
 release = False
 __version__ = "2.0.1.dev0"
-__hash__ = "f7bde69707ac708a758a02d89f14997ee468d1ee"
-__short_hash__ = "f7bde69"
-__date__ = "2024-02-27"
+__hash__ = "5ff9d0de78260383fae0737716879781257155a8"
+__short_hash__ = "5ff9d0d"
+__date__ = "2024-08-30"
--- a/setup.py
+++ b/setup.py
@@ -75,15 +75,16 @@ def glob_paths(pattern):
     for root, dirs, files in os.walk(src_path):
         for file in files:
             if file.endswith(pattern):
-                filepath = os.path.join(str(Path(*Path(root).parts[1:])), file)
+                filepath = os.path.join(root, file)
+                rel_path = os.path.relpath(filepath, src_path)
 
                 try:
-                    out_files.append(filepath.split(f"kivymd{os.sep}")[1])
-                except IndexError:
-                    out_files.append(filepath)
+                    out_files.append(rel_path)
+                except ValueError:
+                    # This might happen if the file is not within src_path
+                    pass
 
     return out_files
-
 
 if __name__ == "__main__":
     # Static strings are in setup.cfg


Novfensec:
I also found kv files missing after doing an iOS build with kivy_ios.

To counter that a friend of mine made a recipe for KivyMD in kivy_ios;

```python
# pure-python package, this can be removed when we'll support any python package
from kivy_ios.toolchain import PythonRecipe, shprint
from os.path import join
import sh
import os


class KivyMDRecipe(PythonRecipe):
    version = "2.0.1"
    url = "https://github.com/kivymd/KivyMD/archive/master.zip"
    depends = ["python"]

    def install(self):
        plat = list(self.platforms_to_build)[0]
        build_dir = self.get_build_dir(plat)
        os.chdir(build_dir)
        hostpython = sh.Command(self.ctx.hostpython)
        build_env = plat.get_env()
        dest_dir = join(self.ctx.dist_dir, "root", "python3")
        build_env['PYTHONPATH'] = self.ctx.site_packages_dir
        shprint(hostpython, "setup.py", "install", "--prefix", dest_dir, _env=build_env)



recipe = KivyMDRecipe()
```

For now we are forcing it to master version by default as its not yet released.
But this builds KivyMD with all kv files on iOS.

There is no pull made for this as it's expected to be resolved in the library itself.

HeaTTheatR:
> I also found kv files missing after doing an iOS build with kivy_ios.
> 
> To counter that a friend of mine made a recipe for KivyMD in kivy_ios;

I've built several iOS apps and never seen the bug you're talking about.

Novfensec:
He told me this (macOS 15). We used a custom fork of kivy_ios to be compatible with buildozer. The error was same as above described. If that's not an issue it's awesome. 👍

Novfensec:
> > I also found kv files missing after doing an iOS build with kivy_ios.
> > 
> > To counter that a friend of mine made a recipe for KivyMD in kivy_ios;
> 
> I've built several iOS apps and never seen the bug you're talking about.

I figured it out. He bundled it explicitly inside the app directory. It only happens in this case. Thanks for clarification.

https://github.com/kivymd/KivyMD/issues/1722

Issue #1735, opened 2024-09-08
MDSegmentButton with Icon raises AttributeError


dev8-de:
### Description of the Bug

if you use the MDSegmentedButtonIcon, the application will exit with an AttributeError:

**AttributeError: 'NoneType' object has no attribute 'parent'**

### Code and Logs
```python

from kivy.lang import Builder

from kivymd.uix.label import MDLabel
from kivymd.uix.boxlayout import MDBoxLayout
from kivymd.uix.segmentedbutton import (
    MDSegmentedButton,
    MDSegmentedButtonItem,
    MDSegmentButtonIcon,
    MDSegmentButtonLabel,
)
from kivymd.app import MDApp

KV = '''
MDScreen:
    md_bg_color: self.theme_cls.backgroundColor

    MDBoxLayout:
        id: box
        orientation: "vertical"
        size_hint_x: .7
        adaptive_height: True
        spacing: "24dp"
        pos_hint: {"center_x": .5, "center_y": .5}
'''


class Example(MDApp):
    def on_start(self):
        for segment_type in ["large", "normal", "medium", "small"]:
            self.root.ids.box.add_widget(
                MDBoxLayout(
                    MDLabel(
                        text=f"Type '{segment_type}'",
                        adaptive_height=True,
                        bold=True,
                        pos_hint={"center_y": 0.5},
                        halign="center",
                    ),
                    MDSegmentedButton(
                        MDSegmentedButtonItem(
                            MDSegmentButtonIcon(icon='cancel'),
                            MDSegmentButtonLabel(
                                text="Songs",
                            ),
                        ),
                        MDSegmentedButtonItem(
                            MDSegmentButtonLabel(
                                text="Albums",
                            ),
                        ),
                        MDSegmentedButtonItem(
                            MDSegmentButtonLabel(
                                text="Podcasts",
                            ),
                        ),
                        type=segment_type,
                    ),
                    orientation="vertical",
                    spacing="12dp",
                    adaptive_height=True,
                )
            )

    def build(self):
        return Builder.load_string(KV)


Example().run()
```

Errormessage:

```
 Traceback (most recent call last):
   File "~/work/project/.venv/lib/python3.12/site-packages/kivy/lang/builder.py", line 245, in create_handler
     return eval(value, idmap), bound_list
            ^^^^^^^^^^^^^^^^^^
   File "MDSegmentedButton", line 55, in <module>
 AttributeError: 'NoneType' object has no attribute 'parent'
 
[...]

PR #1754, opened 2024-12-13
Fix AttributeError on segmented button created in declarative python

alpianon:

See #1735

In segmentedbutton.kv, some code tries to access a MDSegmentedButtonIcon object parent's parent object without first checking if the MDSegmentedButtonIcon object actually has a parent's parent

### Reproducing the problem

See #1735

See #1735

### Description of Changes

In segmentedbutton.kv, we need to check if the MDSegmentButtonIcon object actually has a parent.parent before checking container's parent _label, otherwise you may get the following error: AttributeError: 'NoneType' object has no attribute 'parent'

Now the example code in #1735 works:

immagine

The same example code found in #1735

### DIFF
--- a/kivymd/uix/segmentedbutton/segmentedbutton.kv
+++ b/kivymd/uix/segmentedbutton/segmentedbutton.kv
@@ -52,13 +52,15 @@
         "12dp", \
         0, \
         0 \
-        if self.parent.parent.ids.container.parent._label else \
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
https://github.com/kivymd/KivyMD/commit/9dbf5332257ef2cd60d5bc746773a987a8fec0d2

PR #1756, opened 2025-01-08
Update README.md to include correct install instructions and links

Tear4Pixelation:
The KivyMD package on PyPi is still the old one and does, in fact, not have version 2.0.0 or 2.0.1. I also changed the PyPi links.

Readme.md gives the wrong info on installation

Wrong links and package names

### Reproducing the problem
pip install kivymd==2.0.0

### Description of Changes
Changed links and commands in readme

image

I don't think this is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ project without forking for easier collaboration), have at least one PR
 approved and ask for it. If you contribute regularly to the project the role
 may be offered to you without asking too.
 
-[![PyPI version](https://img.shields.io/pypi/v/kivymd.svg)](https://pypi.org/project/kivymd)
+[![PyPI version](https://img.shields.io/pypi/v/kivymd.svg)](https://pypi.org/project/kivymd2)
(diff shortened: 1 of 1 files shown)


Tear4Pixelation:
The links to pepy are also wrong I just realized

T-Dynamos:
Closing.
There is no intention for `kivymd2`, this version will be published when everything is stable.

Tear4Pixelation:
Huh @T-Dynamos  but why does the package exist? And why is there a reference to a version that does not exist (`kivymd==2.0.0`) but does exist like this 
`kivymd2==2.0.0`

Tear4Pixelation:
I have already seen bug reports for this.

T-Dynamos:
> why does the package exist

This is not official package, someone decided to push it to pypi as it didn't have an release for a long time.

> reference to a version that does not exist (kivymd==2.0.0)

It was changed in `README.md` with intentions to release kivymd with version 2.0.0 back then, but it never happened.

PR #1762, opened 2025-03-03
Added 'get_active_item' to MDNavigationBar

Jersonrn:
Returns the currently active item in the navigation bar.

:return: The active :class:`~kivymd.uix.navigationbar.MDNavigationItem` instance if found, otherwise `None`.

Usage example:

```python
navigation_bar = MDNavigationBar(
        BaseMDNavigationItem(icon="gmail", text="Screen 1", active=True,),
        BaseMDNavigationItem(icon="twitter", text="Screen 2",),
        BaseMDNavigationItem(icon="linkedin", text="Screen 3",),
)
active_item = navigation_bar.get_active_item()
```


HeaTTheatR:
This code is faster and simpler:

```python
def get_active_item(self) -> MDNavigationItem | None:
    """Returns the currently active item in the navigation bar."""

    return next((child for child in self.children if child.active), None)
```

PR #1776, opened 2025-04-30
Fix documentation link in PYPI installation since it's pointing to an…

tepong32:
… old version: kivymd 1.2.0 instead of kivymd2.

I'm a newbie and I just want to explore creating apps using the language and the framework. Apologies if I actually don't know much about forking yet. Thank you.

Older version of the package was linked on the docs and PYPI has a different package reference for the latest version

I tried installing using pip install and "kivymd=2.0.0" was not available.

### Reproducing the problem

```bash
pip install kivymd==2.0.0
```

image

### Description of Changes
Tried installing the package following the instructions and the console said there was no matching version. Latest was 1.2.0 so I went along with it. Only kivymd1.2.0 was installed on my machine so, I tried working on with that and had issues with following docs for kivymd2.0.0. So, I tried searching PYPI for the kivymd keyword and found that there's this kivymd2 package with exactly the same docs with this github repo.

image

```bash
pip install kivymd2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ may be offered to you without asking too.
 ## Installation
 
 ```bash
-pip install kivymd==2.0.0
+pip install kivymd2
 ```
 
 ### Dependencies:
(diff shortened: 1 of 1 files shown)


T-Dynamos:
Sorry, https://pypi.org/project/kivymd2 is not official. And official kivymd 2.0 is yet to release.

PR #1781, opened 2025-05-03
Migrate to new datetime API

emmanuel-ferdman:
The code uses `datetime.datetime.utcfromtimestamp()`, which is deprecated and scheduled for removal in future Python versions:
```python
DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```

- The script attempts to generate a version date from a timestamp using `datetime.utcfromtimestamp()`.
- This triggers a deprecation warning in modern Python versions.

### Description of Changes

This small PR resolves the deprecation warnings by migrating to `datetime.fromtimestamp()`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 import os
 import subprocess
 import sys
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from time import time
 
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
Will

```python
date = datetime.fromtimestamp(epoch, timezone.utc).strftime("%Y-%m-%d")
```

work with later versions of Python?

emmanuel-ferdman:
@HeaTTheatR Yes, this code will work in later versions of Python because:
- `datetime.timezone` was introduced in Python 3.2.
- The `tz` parameter for `datetime.fromtimestamp()` was added in Python 3.6, allowing timezone-aware conversions.
- `datetime.utcfromtimestamp()` is deprecated since Python 3.12.
- `datetime.UTC` (an alias for `timezone.utc`) was introduced in Python 3.11.

In order to support older versions, I kept `timezone.utc` instead of `datetime.UTC`.

HeaTTheatR:
Good!
Thank you for your contribution to the development of the library!

Issue #1817, opened 2025-09-18
MDDropdownMenu: dismiss() should dispatch on_dismiss event instead of calling it directly.


j4ggr:
I am trying to bind actions to the `on_dismiss` event of `MDDropdownMenu`. However, when I call the `menu.dismiss()` method programmatically, the bound callbacks do not get triggered as expected.

The root cause appears to be that inside the `dismiss` method, `on_dismiss` is called directly instead of being dispatched as an event. This bypasses the event system and prevents any bound event handlers from executing.

Currently, the relevant methods look like this:

```python
def on_touch_down(self, touch):
    if not self.menu.collide_point(*touch.pos):
        self.dispatch("on_dismiss")
        return True
    super().on_touch_down(touch)
    return True

def dismiss(self, *args) -> None:
    """Closes the menu."""
    self.on_dismiss()
```

I believe it would be more consistent and intuitive if the `dismiss` method itself dispatched the `on_dismiss` event, and the `on_touch_down` method called `dismiss`. For example:

```python
def on_touch_down(self, touch):
    if not self.menu.collide_point(*touch.pos):
        self.dismiss()  # call dismiss here instead of dispatching directly
        return True
    super().on_touch_down(touch)
    return True

def dismiss(self, *args) -> None:
    """Closes the menu."""
    self.dispatch("on_dismiss")  # dispatch event here
```

This change would ensure that any dismissal of the menu, whether user-initiated or programmatic, consistently triggers the `on_dismiss `event and its bound handlers.

Environment:
- Kivy: v2.3.1
- KivyMD: v2.0.1.dev0
- Python: v3.13.3

HeaTTheatR:
Thank you!
You can create a pull request with these improvements.

HeaTTheatR:
```python
from kivy.lang import Builder
from kivy.metrics import dp

from kivymd.app import MDApp
from kivymd.uix.menu import MDDropdownMenu

KV = '''
MDScreen:
    md_bg_color: self.theme_cls.backgroundColor

    MDButton:
        id: button
        pos_hint: {"center_x": .5, "center_y": .5}
        on_release: app.menu_open()

        MDButtonText:
            text: "Press me"
'''


class Example(MDApp):
    def dismiss(self, *args):
        print(args)

    def menu_open(self):
        menu_items = [
            {
                "text": f"Item {i}",
                "on_release": lambda x=f"Item {i}": self.menu_callback(x),
            } for i in range(5)
        ]
[...]

PR #1818, opened 2025-09-19
Fix MDDropdownMenu.dismiss() to dispatch on_dismiss event properly

j4ggr:
This pull request addresses the issue described in [#1817](https://github.com/kivymd/KivyMD/issues/1817#issue-3430634019), where calling `dismiss()` on `MDDropdownMenu` does not trigger the `on_dismiss` event handlers because `on_dismiss` is called directly instead of dispatched.

Changes made:
- Modified the `dismiss()` method to dispatch the `on_dismiss` event.
- Updated `on_touch_down` to call `dismiss()` instead of dispatching `on_dismiss` directly.

This ensures consistent and correct event dispatching whether the menu is dismissed by user interaction or programmatically.


HeaTTheatR:
https://github.com/kivymd/KivyMD/issues/1817

PR #1841, opened 2025-12-26
Allows number verification on MDTextfield

Jo-e7:
This adds number verification functionality to MDTextField. Here is a demo of the functionality:

https://github.com/user-attachments/assets/54b9e9f5-6786-4629-abb9-00a7c1720458

The code used for the demo was:

```
from kivy.lang import Builder

from kivymd.app import MDApp

KV = '''
MDScreen:
    md_bg_color: app.theme_cls.surfaceColor

    MDTextField:
        mode: "outlined"
        size_hint_x: None
        width: "240dp"
        validator: "number"
        MDTextFieldLeadingIcon:
            icon: "numeric"
        MDTextFieldHintText:
            text: "Insert a number"

'''


class Number_validator(MDApp):
    def build(self):


Number_validator().run()
```

### DIFF
--- a/kivymd/uix/textfield/textfield.py
+++ b/kivymd/uix/textfield/textfield.py
@@ -457,6 +457,13 @@ class Validator:
     :attr:`date_format` is an :class:`~kivy.properties.OptionProperty`
     and defaults to `None`.
     """
+    def is_number_valid(self, text: str) -> bool:
+        """Checks the validity of a number."""
+        for i in text:
+            if not i.isdigit():
+                return True
+                break
+        return False
 
     def is_email_valid(self, text: str) -> bool:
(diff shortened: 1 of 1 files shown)


HeaTTheatR:
The code contains several problems:

- A logic error: the function returns `True` on the first non-digit, and `False` only if all characters are digits. This is the opposite of the typical behavior of `is_number_valid` (which usually returns `True` for valid numbers).
- An redundant `break`: the return already breaks the loop.
- An incorrect name: the function returns `True` on an error.

Easier:

```python
def is_number_valid(self, text: str) -> bool:
    """Checks if the text contains only digits."""

    return text.isdigit()
```

